PR #1, opened 2018-01-23
Fixes example ant build script dependency and adds 'onDraw()' to start engine

nargetdev:
Fixes the Example ant build script dependency and adds 'lx.engine.onDraw()' which seems to be necessary to start the engine thread.  Finally gives instructions on how to actually run this thing headless.


mcslee:
Thanks dude, some of these issues were from other bugs I had failed to clean up. I ended up fixing them in a slightly different way - https://github.com/heronarts/LX/commit/2821236599170c3b3de6b01da55b6dabae91f01b - this should fix all the concerns brought up here.

PR #2, opened 2018-02-16
Fixed e1.31 header index values.

jkbelcher:
Header indexes 117 & 118 were mislabeled as 116 & 117 in the constructor.  With the wrong indexes the e1.31 streaming protocol didn't work.  With this fix it does!


mcslee:
Whoops! Nice catch - the E1.31 controller I had used in the past must have just been ignoring those fields. Thanks for fixing.

jkbelcher:
Sure thing.  It was a PixLite controller that didn't like it.  Works now with the PixLite 16 MkII.

PR #3, opened 2018-03-09
Fix sACN datagram source name padding index

pidge:
It was overwriting the whole field by starting from the beginning rather than after the last name character


mcslee:
Whoops, nice catch. Sorry about that and thanks for fixing!

PR #4, opened 2018-03-23
example outputting to Tenere controller

nargetdev:
I decreased the size of the model (derived from the grid_model(30,30) instantiation)  because I see an index out of bounds exception when it's bigger.  Not sure what that's about.  Also I made the `fixtureToIndices` method of LXFixture public so I could extract the indices when needed from outside.  I think it would probably be a better idea to expose a method of LXFixture to get the indices.


examples/LXHeadless/src/heronarts/lx/headless/LXHeadless.java
```diff
@@ -22,10 +22,7 @@
 import heronarts.lx.LXPattern;
 import heronarts.lx.model.GridModel;
 import heronarts.lx.model.LXModel;
-import heronarts.lx.output.ArtNetDatagram;
-import heronarts.lx.output.FadecandyOutput;
-import heronarts.lx.output.LXDatagramOutput;
-import heronarts.lx.output.OPCOutput;
+import heronarts.lx.output.*;
```
mcslee, line 26: Are you using an IDE? Better to specifically enumerate only the dependencies that are used. Eclipse or IntelliJ should both do this...

examples/LXHeadless/src/heronarts/lx/headless/LXHeadless.java
```diff
@@ -35,16 +32,16 @@
 
   public static LXModel buildModel() {
     // TODO: implement code that loads and builds your model here
-    return new GridModel(30, 30);
+    return new GridModel(10,17);
```
mcslee: Before changing this, I'd love to see what the IndexOutOfBounds exception was that was being caused by 30, 30. That's concerning, implies that some size limit was being improperly hardcoded elsewhere...

examples/LXHeadless/src/heronarts/lx/headless/LXHeadless.java
```diff
@@ -55,14 +52,47 @@ public static void addOPCOutput(LX lx) throws Exception {
...
     try {
       LXModel model = buildModel();
       LX lx = new LX(model);
 
+
+      addTenereOutput(lx, "10.200.1.64");
+      addTenereOutput(lx, "10.200.1.65");
+
+      // target some OPC servers
+      String[] controller_ips = { "10.200.1.146", "10.200.1.195", "10.200.1.197"};
```
mcslee: Minor naming convention thing, but should keep the style conventions standard in here. lower_caps_underscore shouldn't be used here, rather should be camelCase.

I'm also not sure these specific IP addresses and Tenere-specific output classes ought to go in the base project, happy to let this slide but let's try to keep project-specific stuff to a minimum in here.

mcslee (changes requested):
Also made a change here so there's a public utility function for getting a list of point indices from any LXFixture, which is LXFixture.Utils.getIndices(LXFixture fixture)

https://github.com/heronarts/LX/commit/01e6a2654c2cd8fc10810a27d0919a0927d64e16

mcslee:
Yo - so it seems like the merge action here has gotten a little weird. Have you rebased your branch against heronarts:master? It looks like all the diffs that I created since are being double-included in the change-set here, rather than it being re-based against an update.

Do you want to do:
git pull --rebase

In your branch from the point that had your commits, and then generate a new clean pull request based upon that?

nargetdev:
Ok yea, I felt like there might be some weirdness – I’ll try again when I have a second.  There should only be like 15 lines of code added to the example, and then the TenereController.java class

PR #6, opened 2018-06-24
Helpful functions for subclasses

cinderblock:
I was writing a subclass and found myself recreating functionality so why not expose this useful functionality.


mcslee:
Just FYI - https://github.com/heronarts/LX/commit/bacc5b26f8debed0f4ff09312ce5dce1c5ee124f - added bounds checking and changed naming to "set" instead of "write" - the Java convention is that write() or send() typically actually pushes data to an output stream. Whereas this is just updating the buffer and not actually sending it.

mcslee:
And updated RadiaLumia JAR here: https://github.com/FoldHaus/radialumia/commit/607c985acab68bbf9412892f3f926de91bf520b6

PR #11, opened 2018-08-07
Bug fix: don't set value on FunctionalParameter.

jkbelcher:


mcslee:
Yep that'll do it. Might make sense to also modify the save function to not even write the parameter into the file in the first place - but this is def the more important side.

Issue #17, opened 2018-12-15
Request: Convert build to Maven or Gradle


drichelson:
This makes for zero effort import into IDEs and makes for easy publishing of jars which makes importing LX into a project much easier. I'm happy to do this work if you will consider merging it.

mcslee:
Hey Dan - yep I'd be happy to merge this. I would still want to keep the simple ant build scripts around, but totally happy to have this also integrate with the more modern build systems. I haven't managed to prioritize this, if you are familiar and have time to quickly whip it up that'd be great.

nargetdev:
Did this ever happen?  Would be happy to take a crack at it if no.

mcslee:
Yup - everything is Mavenized, and releases are published to Maven Central.
https://search.maven.org/search?q=g:com.heronarts

How you doing @nargetdev? Got LX projects in the works these days?

PR #18, opened 2018-12-17
Mavenize project.

drichelson:
Addresses #17.
Add maven support. This involved moving all source files into standard Maven locations. This was done via `git mv` command so history remains intact. Everything that was possible with ant before should doable with Maven, if not please let me know what is missing.


mcslee:
Thanks for that. Quick question - what's the recommended process for version updates / releases?

I assume master keeps working with version as: "0.1.3-SNAPSHOT"

When that's finished, shall I just make a commit to change it to "0.1.3" - then cut a release and a tag, and then immediately make another commit to change version to "0.1.4-SNAPSHOT"

Let me know if that'll be the best way to make it play nice with others...

drichelson:
@mcslee The process you describe is exactly how I would do it. It is also exactly how the Maven release plugin works which may be a good option if you are releasing often or will have a CI system perform a release. Otherwise I'd stick with the manual approach.

PR #19, opened 2018-12-29
Adjust Maven project structure and settings.

drichelson:
Addresses #17 (again)
This improves the naming of artifacts in the local Maven repo when running `mvn install`
Before the artifacts were:
`.../heronarts/core/...`
`.../heronarts/headless-example/...`
Now they are:
`.../heronarts/LX/...`
`.../heronarts/LX-headless-example/...`

I've also now made the thin jar actually thin so it no longer includes the dependencies.

Both of these changes will make for more developer-friendly artifacts should you decide to publish jars to a Maven repository.


mcslee:
Great - thanks!

PR #21, opened 2019-02-28
DJM900nxs2 midi surface

jkbelcher:
Pioneer DJM900-NXS2 midi surface, first attempt.  All MIDI controls values are identified but only EQ, color levels, and basic faders are mapped to parameters.

### DIFF
--- a/LX/src/main/java/heronarts/lx/midi/surface/DJM900nxs2.java
+++ b/LX/src/main/java/heronarts/lx/midi/surface/DJM900nxs2.java
@@ -1,0 +1,590 @@
+package heronarts.lx.midi.surface;
+
+import heronarts.lx.LX;
+import heronarts.lx.midi.LXMidiInput;
+import heronarts.lx.midi.LXMidiOutput;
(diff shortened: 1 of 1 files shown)


mcslee:
Sorry took me a minute to get to this. Looks good, FYI I'll change a few things after I accept it and shoot you a pointer to what I've done. The parameters shouldn't be CompoundParameter because nothing else should be modulating these... they should always be raw values used to modulate other stuff (it's not really a problem for them to be compound, just a matter of semantics).

Also gonna introduce an enum for the channel numbers because the 1-5 for loops and channel-1 stuff  makes me feel a touch icky.

Thx for mapping this all out!

mcslee:
FYI - changes. I haven't tested since I don't have a DJM900. Huge apologies if I broke anything here. But I think this should be a little more robust to indexing-bugs with the Channel thing spelling out more explicitly what's a 1-4 "user-facing" number and what's an array index.
https://github.com/heronarts/LX/commit/c2b73b334ab9930514580474ffd3f9aaf99028ce

jkbelcher:
Excellent.

PR #22, opened 2019-03-07
Allow output brightness to be modulation target

jkbelcher:
Can we change this to allow the master fader / output brightness to be a modulation target?


mcslee:
Yeah I suppose there's no reason not to. This used to be more costly because the LXOutput class was doing a bunch of scaling when brightness was not at max, but it all uses a lookup table now, so I think this is fine. Thanks!

PR #28, opened 2019-04-08
fix javadoc link refs

tracyscott:
javadoc failure aborts mvn package.  Another option is to set maven javadoc plugin parameter failOnError to false if mvn package/javadoc isn't a hard pre-commit requirement due to differing workflows.


mcslee:
Fixed by https://github.com/heronarts/LX/commit/fd5c19a5adafa9cc18b81803340e8624c02f4e8e - thanks though!

PR #32, opened 2020-02-26
Snapshots

jkbelcher:
Intro version of snapshots.  I elected to use the name snapshot rather than scene to avoid confusion with clips.  Handles channel/pattern add, move, delete.  Saves and loads from project file.  Individual snapshots can be launched with OSC commands.  Does not currently function with the Undo framework.  Also does not restore values if a channel/pattern is deleted and then the delete is undone.  Will send a separate pull request with accompanying UI.


mcslee:
Nice - this is definitely on the right track. There's enough in here that I'd want to change that I'd rather not merge this now if you're cool with that. Can get it in a branch if that makes it more workable for you (or just keep using your remote branch). I'll leave a review with various comments on some of the things I'd modify here. No pressure to do that if this is working fine for you right now - as to get this into the main branch there'll be deeper changes needed to be compatible with the Undo engine, which are probably not worth your time on a Treefort prep schedule.

LX/src/main/java/heronarts/lx/LXEngine.java
```diff
@@ -93,6 +93,8 @@
 
   public final LXScriptEngine script;
 
+  public final LXSnapshotController snapshotController;
```
mcslee: There's enough here that I would put everything inside a package heronarts.lx.snapshot. would also call this LXSnapshotEngine for naming parity, and just call this variable "snapshots" so it's lx.engine.snapshots similar to lx.engine.modulation, lx.engine.audio etc.

LX/src/main/java/heronarts/lx/LXEngine.java
```diff
@@ -482,6 +484,10 @@ public void onParameterChanged(LXParameter p) {
       });
     }
 
+    // Snapshots
+    addChild("snapshot", this.snapshotController = new LXSnapshotController(lx));
```
mcslee: Possibly "snapshots"

LX/src/main/java/heronarts/lx/LXSnapshot.java
```diff
@@ -0,0 +1,243 @@
...
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+import heronarts.lx.modulator.LXModulator;
+import heronarts.lx.parameter.CompoundParameter;
+import heronarts.lx.parameter.LXParameter;
+
+public class LXSnapshot extends LXComponent {
```
mcslee: I think the user should be able to rename these like patterns, etc. Get this with:

implements LXComponent.Renamable

LX/src/main/java/heronarts/lx/LXSnapshot.java
```diff
@@ -0,0 +1,243 @@
...
+
+  public Integer snapshotID;
```
mcslee: private int index;

Shouldn't be public, and no need for heavy Integer type. I'd look at LXModulator.java and follow the naming conventions for setIndex() / getIndex() etc.

LX/src/main/java/heronarts/lx/LXSnapshot.java
```diff
@@ -0,0 +1,243 @@
...
+
+  public final List<LXSnapshotValue> snapshotValues = new ArrayList<LXSnapshotValue>();
```
mcslee: Note the pattern used in other LX classes for providing public visibility to a mutable list

  private final List<LXCompoundModulation> mutableModulations = new ArrayList<LXCompoundModulation>();
  public final List<LXCompoundModulation> modulations = Collections.unmodifiableList(this.mutableModulations);

LX/src/main/java/heronarts/lx/LXSnapshotActivePattern.java
```diff
@@ -0,0 +1,39 @@
+package heronarts.lx;
+
+import com.google.gson.JsonObject;
+
+public class LXSnapshotActivePattern extends LXComponent {
```
mcslee: I might consider making this a private inner class of LXSnapshot. And I'm not really sure if this ought to extend LXComponent. Would just be inside LXSnapshot and look like:

private class ActivePattern {
}

LX/src/main/java/heronarts/lx/LXSnapshotActivePattern.java
```diff
@@ -0,0 +1,39 @@
...
+
+  public LXChannel channel;
```
mcslee: These raw references are fine for a quick-and-dirty solution now. But for a long-term version they shouldn't be public and mutable.

You can take a look at how the Undo engine handles this stuff via the ComponentReference and ParameterReference classes. I will eventually bubble those up into a more public location in the API.

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+import java.util.List;
+import java.util.TreeMap;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+import heronarts.lx.osc.OscMessage;
+
+public class LXSnapshotController extends LXComponent {
```
mcslee: LXSnapshotEngine

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+
+  private Integer nextShapshotID = 1;
+
+  private final List<LXSnapshot> snapshots = new ArrayList<LXSnapshot>();
+  private final TreeMap<Integer, LXSnapshot> snapshotsDict = new TreeMap<Integer, LXSnapshot>();
```
mcslee: Should probably just be the array with reordering support, similar to how modulators are handled in the LXModulatorComponent.

This way you can also use addArray() for the component saving.

See:
https://github.com/heronarts/LX/blob/master/LX/src/main/java/heronarts/lx/LXModulatorComponent.java#L48

Scanning the array for lookup-by-index is fine, that's super-cheap and no risk of data inconsistency maintaining two containers of the same objects.

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+  public interface SnapshotListener {
+    public void snapshotAdded(LXSnapshotController controller, LXSnapshot snapshot);
+    public void snapshotRemoved(LXSnapshotController controller, LXSnapshot snapshot);
+  }
+
+  private final List<SnapshotListener> snapshotListeners = new ArrayList<SnapshotListener>();
+
+  public LXSnapshotController(LX lx) {
+    super(lx);
+    this.label.setDescription("Snapshots");
```
mcslee: I think you might mean setValue() here? But better to use a different super() constructor:

super(lx, "Snapshots");

That sets default label.

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+    return _addSnapshot(snapshot);
+  }
+
+  public LXSnapshot addSnapshot() {
+    LXSnapshot snapshot = _createNewSnapshot();
+    _addSnapshot(snapshot);
+    return snapshot;
+  }
+
+  public void runSnapshot(int snapshotID) {
```
mcslee: I get the intuitive reason for "run", but feel like this would be more consistent as something like loadSnapshot().

Also wonder whether this should be invoked by integer index on the controller object, vs. being a direct "load" method on the LXSnapshot class itself.

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+    if (snapshotsDict.containsKey(snapshotID)) {
+      snapshotsDict.get(snapshotID).run();
+    } else {
+      System.err.println("Unknown snapshotID, cannot run snapshot.");
```
mcslee: e.g. would be no need for this if the method was on the LXSnapshot object itself

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+
+  public LXSnapshotController removeSnapshot(LXSnapshot snapshot) {
+    this.snapshotsDict.remove(snapshot.snapshotID);
+    this.snapshots.remove(snapshot);
+    for (SnapshotListener snapshotListener : this.snapshotListeners) {
+      snapshotListener.snapshotRemoved(this, snapshot);
+    }
+    snapshot.dispose();
+
+    if (this.snapshots.size()==0) {
```
mcslee: I think it's probably better to always reindex the snapshots here for consistency. Maintaining gaps in the indices seems likely to cause confusion. See removeModulator() on LXModulatorComponent

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+import heronarts.lx.osc.OscMessage;
+
+public class LXSnapshotController extends LXComponent {
+
+  private Integer nextShapshotID = 1;
+
+  private final List<LXSnapshot> snapshots = new ArrayList<LXSnapshot>();
+  private final TreeMap<Integer, LXSnapshot> snapshotsDict = new TreeMap<Integer, LXSnapshot>();
+
+  public interface SnapshotListener {
```
mcslee: Fine to just call this Listener, so the full class name is LXSnapshotEngine.Listener

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+    this.snapshotListeners.add(listener);
+    return this;
+  }
+
+  public LXSnapshotController removeSnapshotListener(SnapshotListener listener) {
+    this.snapshotListeners.remove(listener);
+    return this;
+  }
+
+  public LXSnapshotController removeLinks(LXComponent component) {
```
mcslee: For naming consistency might call this removeReferences()

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+      } else {
+        System.out.println("Invalid snapshot ID received over OSC: " + snapshotID);
+      }
+    } catch (NumberFormatException exception) {
+      System.err.println("Invalid snapshot ID received over OSC: " + path);
+    }
+    return true;
+  }
+
+  private static final String KEY_NEXTSNAPSHOTID = "nextSnapshotID";
```
mcslee: Shouldn't need to store this if stick to flat array

LX/src/main/java/heronarts/lx/LXSnapshotController.java
```diff
@@ -0,0 +1,161 @@
...
+  private static final String KEY_SNAPSHOTS = "snapshots";
+
+  @Override
+  public void save(LX lx, JsonObject obj) {
+    super.save(lx, obj);
+    obj.addProperty(KEY_NEXTSNAPSHOTID, this.nextShapshotID);
+    obj.add(KEY_SNAPSHOTS, LXSerializable.Utils.toArray(lx, this.snapshots));
```
mcslee: Not needed if use addArray() similar to LXModulatorComponent does with modulators

LX/src/main/java/heronarts/lx/LXSnapshotValue.java
```diff
@@ -0,0 +1,64 @@
+package heronarts.lx;
+
+import com.google.gson.JsonObject;
+
+import heronarts.lx.parameter.LXParameter;
+
+public class LXSnapshotValue extends LXComponent {
+
+  public Boolean isEnabled;
```
mcslee: Should be a BooleanParameter if this would be toggle-able in the UI and saved as a setting.

LX/src/main/java/heronarts/lx/LXSnapshotValue.java
```diff
@@ -0,0 +1,64 @@
...
+
+  public double value;
```
mcslee: These should be private if they're mutable...

LX/src/main/java/heronarts/lx/LXSnapshotValue.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  public LXSnapshotValue(LX lx, LXComponent scope, JsonObject obj) {
+    super(lx);
+    this.isEnabled = obj.get(KEY_ENABLED).getAsBoolean();
+    this.value = obj.get(KEY_VALUE).getAsDouble();
+    this.parameter = getParameter(lx, scope, obj.getAsJsonObject(KEY_PARAMETER));
+  }
+
+  public static LXParameter getParameter(LX lx, LXComponent scope, JsonObject obj) {
```
mcslee: Looks redundant with LXParameterModulation.getParameter()... this should probably be pushed up to being a more generic static method on the LXComponent class...

LX/src/main/java/heronarts/lx/LXSnapshotValue.java
```diff
@@ -0,0 +1,64 @@
...
+  protected static final String KEY_VALUE = "value";
+  protected static final String KEY_PARAMETER = "parameter";
+
+  @Override
+  public void save(LX lx, JsonObject obj) {
+    super.save(lx, obj);
+    obj.addProperty(KEY_ENABLED, this.isEnabled);
+    obj.addProperty(KEY_VALUE, this.value);
+
+    JsonObject parameterObj = new JsonObject();
```
mcslee: This is fine, though maybe opportunity for some code reuse or a helper in the LXSerializable class. Since this is done very similarly in the LXParameterModulation class

mcslee (commented):
No pressure to address all these things, just noting stuff.

jkbelcher:
Agreed, this version is definitely not flushed out yet.  It's enough to be functional for our show this weekend and possibly for Treefort too depending on time.  Figured I would at least get your eyes on it before going much further.

It's enough for me to have this in my own branch.  Or is it easier to discuss changes if it's in a heronarts:snapshots branch?

LX/src/main/java/heronarts/lx/LXEngine.java
jkbelcher: Thanks.  Definitely can use help on LX-style organization and naming conventions.

LX/src/main/java/heronarts/lx/LXSnapshot.java
jkbelcher: Renaming actually works from the ui, not sure why if it should require this interface.

LX/src/main/java/heronarts/lx/LXSnapshotActivePattern.java
jkbelcher: Agreed.  It was a more complicated class at one point before I changed it to saving just the indexes.

LX/src/main/java/heronarts/lx/LXSnapshotController.java
jkbelcher: Sounds good.

LX/src/main/java/heronarts/lx/LXSnapshotController.java
jkbelcher: I think snapshotIDs (or snapshot indexes) need to remain constant once they're created, because I'm setting up OSC links to them from an external program.  I.E., "load snapshot 7 when I press this button" by sending OSC command \lx\snapshot\7.  Wouldn't want to reindex for instance if I delete snapshot 2 because then my external program is out of sync.  My reasoning was that the one time it is acceptable to start the snapshotIDs/indexes over is if the entire list is cleared.  Then it's nice to reset to 1 to keep things clean.  This was part of my reason for using the name snapshotID instead of index, because index implies sequential ordering.  The function here is a unique identifier that remains constant with the snapshot instance.

LX/src/main/java/heronarts/lx/LXSnapshotValue.java
jkbelcher: Yes.

jkbelcher:
Feedback points look good.  I'll implement them when time permits.

mcslee:
> Or is it easier to discuss changes if it's in a heronarts:snapshots branch?

Nah easier to just stick in your own branch, that way you can just keep updating this thread with your changes, keep it in one place, no merging mess.

LX/src/main/java/heronarts/lx/LXSnapshot.java
mcslee: https://github.com/heronarts/LX/blob/master/LX/src/main/java/heronarts/lx/LXComponent.java#L660

It won't be properly restored on the next run. This is to prevent people from going rogue and renaming things like the core Engine, Modulation Engines, Audio Engine, etc.

jkbelcher:
Do you have any further thoughts on this issue of indexing the internal list of snapshots versus keeping the snapshot identifiers constant?  The issue is that I create a snapshot and set up an accompanying OSC trigger from an external program to launch, for example, /lx/snapshots/37.  If we used only indexes for identifiers, and I were to delete snapshot 4, all of the indexes would get adjusted and my external program would be out of sync.  If you really want to stick with just indexes, I can avoid this problem by never deleting a snapshot from early in the list.  Although I feel like my scenario is a common use case when triggering from a lighting console or other external software.  I also think it might be common to set up cues from a lighting console but then reorder the list into the order they will be played.

My inclination would be to change the list of snapshots to be index-based and therefore managed and reorderable like other LX collections, but to have an extra identifier on the LXSnapshot class which remains constant through list reordering, and which is used by OSC commands.

mcslee:
> My inclination would be to change the list of snapshots to be index-based and therefore managed and reorderable like other LX collections, but to have an extra identifier on the LXSnapshot class which remains constant through list reordering, and which is used by OSC commands.

Yep that's what I'd recommend. This is how other LX objects are addressed via OSC, e.g. patterns and effects. They can either be looked up by their ordered index (which is maintained with no gaps) - or by string identifier. OSC respects both. The identifier labels stay the same across moves, the indices can change.

Can see some reference code for this with patterns here:
https://github.com/heronarts/LX/blob/master/LX/src/main/java/heronarts/lx/LXChannel.java#L306

Effects here:
https://github.com/heronarts/LX/blob/master/LX/src/main/java/heronarts/lx/LXBus.java#L129

Basically, if the OSC path is numeric, match on index, otherwise try to match on unique label identifier.

jkbelcher:
Incremental improvements as discussed plus a few others.  Not yet ready for main branch merge, needs a few features including undo support.

Changes as discussed:
	-Moved classes to package heronarts.lx.snapshot
	-Renamed LXSnapshotController to LXSnapshotEngine
	-Renamed LXEngine child from "snapshot" to "snapshots"
	-Added Renamable interface to LXSnapshot
	-OSC references to snapshots can use an integer for index or non-integer for label, like other LX objects
	-Renamed LXSnapshot.run() to load()
	-Changed LXSnapshotEngine super() constructor
	-Renamed LXSnapshotEngine.SnapshotListener to Listener
	-Renamed LXSnapshotEngine.removeLinks() to removeReferences()
	-Removed nextShapshotID property from LXSnapshotEngine
	-Used addArray() with LXSnapshotEngine child object snapshots
	-Changed LXSnapshotValue.isEnabled to BooleanParameter
	-Changed LXSnapshotValue.value visibility to private
	-Changed LXSnapshot.snapshotID to index, LX-style
	-Allow snapshot moving within collection
	
Additional changes:
	-Added isLoadingSnapshot flag and accessor method on SnapshotEngine
	-Snapshots include effects
	-Can "update" a snapshot (at LX level, not yet UI level)
	-Default snapshot names use an auto-incrementing naming scheme

mcslee:
FYI - just got around to building this out with the full undo integration etc. Some more features to come soon, I'm going to make these smoothly interpolatable, add some contextual menus to control how they operate (e.g. like the Lightroom develop settings kind of idea).

But the core components are all in here, should be working:
https://github.com/heronarts/LX/commit/c6fba6d59c3ccb273eabb07a2f5b18cd8efba78c

I elected to call the sub-components of a Snapshot a View.

Basic UI dropped in for now, and I'm going to build that out a bit soon.

mcslee:
Included your name in the author comments on those new files btw.

PR #34, opened 2021-09-08
Add MidiFighterTwister midi controller

jkbelcher:
The Midi Fighter Twister from DJ Tech Tools is a great midi controller for LX.  Sixteen limitless knobs with LED indicators and hardware virtualization of 4 "banks" so the controller presents to software as having 64 knobs.  Each knob includes a push-click function which is very intuitive when mapped to BooleanParameters.  When mapped to a DiscreteParameter currently the push-click function increments the parameter by one; this is useful for clicking through enum values.  Side buttons on the controller allow movement between virtual banks and adjustment of controller target within LX.

This is a working version but still is a little rough around the edges.  Known issues: 1) listeners to mapped parameters are stranded on program exit, 2) when running two of these controllers not all functionality works properly on the second controller.

### DIFF
--- a/src/main/java/heronarts/lx/midi/LXMidiEngine.java
+++ b/src/main/java/heronarts/lx/midi/LXMidiEngine.java
@@ -28,6 +28,7 @@ import heronarts.lx.midi.surface.APC40;
 import heronarts.lx.midi.surface.APC40Mk2;
 import heronarts.lx.midi.surface.DJM900nxs2;
 import heronarts.lx.midi.surface.LXMidiSurface;
+import heronarts.lx.midi.surface.MidiFighterTwister;
 import heronarts.lx.mixer.LXAbstractChannel;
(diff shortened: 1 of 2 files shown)


mcslee:
Nice! Yeah great little device, I've got one lying around somewhere in SF, never used it enough, but super flexible. Just heading out for a week or so visiting Cornwall with fam here, but I will take a close look at this one when I'm back in a week or so and get it merged up.

Also - that whole outputs refactor is finally getting *close* to the top of my TODO list. Will eventually get there I swear!

mcslee:
Generally good - just dropped some cleanup comments on the commit, think the ChannelListener stuff ought to get cleaned out before landing this.

jkbelcher:
Changes submitted.  Still have the boolean isRegistered, to avoid the issue where surface is enabled but channel is null due to no channels.

PR #37, opened 2021-09-16
Find correct midi output when using two devices of same name.

jkbelcher:
A proposed direction for this problem.  Uses a hidden unique deviceID to match midi outputs to midi inputs.  This works on a fresh project file.  However it crashes when loading from a saved file.  Could use your input on whether you like this direction before taking it any further.

The unique id remains constant across multiple runs of LX but I suspect it is only unique to the running instance of OSX. In which case if we write the uniqueId to a saved project file, we need to handle the ids having changes after a computer restart. Suspect the solution here would be to pair the saved instances sequentially with current instances of the same midi device. Could use your assistance with this one for sake of efficiency.

### DIFF
--- a/src/main/java/heronarts/lx/midi/LXMidiDevice.java
+++ b/src/main/java/heronarts/lx/midi/LXMidiDevice.java
@@ -106,6 +106,15 @@ public abstract class LXMidiDevice {
   }
 
   /**
+   * Get the unique name of the device.
+   *
+   * @return Unique device name
+   */
+  public String getNameUnique() {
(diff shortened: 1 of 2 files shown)


mcslee:
Hrm, yeah this will work but only on Mac where the CoreMidi4J library is there. I'll mull it a bit. I think the right solution is to send out a handshake by pinging a MIDI identity request on the output ports, and waiting to see which input port responds: http://midi.teragonaudio.com/tech/midispec/identity.htm

Might require a bit of clunky/cludgy timing code. But can perhaps be limited to doing this check only in the case where there are multiple devices with the same matching name.

I'll sleep on this and mull a touch more.

PR #41, opened 2022-01-11
Bug fix: byte order W sending minimum rgb value instead of composite.

jkbelcher:
Looks like a copy and paste error from the RGBW section below it.  This was sending the minimum value of (r,g,b) to w which on a fully saturated color was often zero.  This fix sets it to the average, although some sources online say it should be an uneven weighting formula such as (W = R^gamma * .2126 + G^gamma * .7152 + B^gamma * .0722).

### DIFF
--- a/src/main/java/heronarts/lx/output/LXBufferOutput.java
+++ b/src/main/java/heronarts/lx/output/LXBufferOutput.java
@@ -19,6 +19,7 @@
 package heronarts.lx.output;
 
 import heronarts.lx.LX;
+import heronarts.lx.utils.LXUtils;
 
 public abstract class LXBufferOutput extends LXOutput {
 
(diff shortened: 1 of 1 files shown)


src/main/java/heronarts/lx/output/LXBufferOutput.java
```diff
@@ -127,7 +128,7 @@ protected LXBufferOutput updateDataBuffer(int[] colors, byte[][] glut, double br
             int r = ((color >> 16) & 0xff);
             int g = ((color >> 8) & 0xff);
             int b = (color & 0xff);
-            int w = (r < g) ? ((r < b) ? r : b) : ((g < b) ? g : b);
+            int w = LXUtils.constrain((r + b + g) / 3, 0, 255);
```
mcslee: Drop the LXUtils.constrain call here... there is no way that this value can be under 0 or over 255, it won't matter on most projects but that's potentially a ton of extra function calls doing nothing for huge pixel count projects.

The only thing I might consider instead is a Math.round(), but I am inclined to just leave it off. If people are sending W-only then they should generally be animating patterns in gray, in which case r=g=b anyways and there will be no rounding errors.

Nuke the function call and we're good to go here - thank you for catching this, definitely an oversight on my part!

mcslee (changes requested):
Let's just kill the constrain() call which won't ever do anything, thanks for catching!

src/main/java/heronarts/lx/output/LXBufferOutput.java
jkbelcher: Good lesson there.  Burning CPU cycles when I should have checked to see if it was necessary.

PR #42, opened 2022-01-18
Combining bipolar and exponent

jkbelcher:
Bipolar and exponent are both great features on CompoundParameter.  However when you combine them you get an asymmetrical knob which visually defeats the purpose of bipolar.  But it would still be nice to have both, so on a bipolar knob one can increase resolution on either side of center without sacrificing range.

What would you think about changing the calculation on bipolar+exponent to have the center of the range be halfway between the min&max and have the log scale mirrored on either side?

I started to work it up here... The math seems right but it's not behaving correctly. Worth getting your opinion on this change before chasing it further.

### DIFF
--- a/src/main/java/heronarts/lx/parameter/BoundedParameter.java
+++ b/src/main/java/heronarts/lx/parameter/BoundedParameter.java
@@ -57,13 +57,25 @@ public class BoundedParameter extends LXListenableNormalizedParameter {
     }
 
     public double getNormalized(double value, double exponent) {
+      return getNormalized(value, exponent, Polarity.UNIPOLAR);
+    }
+
+    public double getNormalized(double value, double exponent, LXParameter.Polarity polarity) {
       if (this.v0 == this.v1) {
         return 0;
       }
       value = constrain(value);
       double normalized = (value - this.v0) / this.vRange;
       if (exponent != 1) {
-        normalized = Math.pow(normalized, 1 / exponent);
+        if (polarity == Polarity.UNIPOLAR) {
+          normalized = Math.pow(normalized, 1 / exponent);
+        } else {
+          if (normalized < .5) {
+            normalized = (1 - Math.pow(1 - (normalized * 2), 1/exponent)) / 2;
+          } else {
+            normalized = .5 + (Math.pow((normalized - .5) * 2, 1/exponent) / 2);
+          }
+        }
       }
       return normalized;
     }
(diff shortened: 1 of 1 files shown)


mcslee:
Ah, interesting idea. I can see the appeal here for that kind of knob behavior, but I think this would have to be specified as a separate option. As is - this would definitely break/modify the behavior of existing parameters, which may be intentional.

I am not sure it is worth the additional complication. It is also of course always possible in pattern/effect code to scale/mutate parameter values appropriately. The exponent thing is a nice utility for making knobs more usable in generic cases, but there's always plenty of room for pattern/effect developers to put the scaling right in code.

e.g.

float rawValue = knob.getValuef();
float scaledValue = Math.pow(rawValue, 2.4) - offset*whatever + etc...

That said, I do particularly see the appeal here for bipolar knobs where the mid-point is 0. But would just have to mull over how to cleanly ask for that without tacking too much on to the parameter class interfaces...

jkbelcher:
Thinking out loud for the sake of concept organizing:
The djm mixer has adjustments for channel fader curves and crossfader curves.  This would be like adding another curve option to a parameter.  Previously there was linear and log, this would be the addition of mirrored log.

I've been using midpoint-zero parameters regularly on the MFT.  The push-click to reset snaps them back to center which makes it fast to clear a setting.  With the knobs on the APC40mkII it was less appealing to use midpoint-zero / bipolar parameters because it was harder to return them to the Off position.

mcslee:
You were pretty close here, I think the crossfader curve metaphor is the right one, I've added that. There were some more calls that needed to be updated in CompoundParameter. I've also added 4 different curves, as there are different ways of biasing.

The NORMAL is 0->1, I made a REVERSE that operates in the opposite direction (e.g. exponential but from 1->0).

Then there are two symmetric curves around the midpoint BIAS_CENTER, and BIAS_OUTER

Say you have -100 -> 100, BIAS_CENTER has a flat response at 0, with exponential curves accelerating away from center.

BIAS_OUTER hits 0 at the center, but the curve is STEEPEST at this point, flattening out at is reaches the extremes.

BIAS_CENTER is *probably* the more useful one for most cases, but not necessarily.

Take these for a spin, let me know if they work out for you?

https://github.com/heronarts/LX/commit/249a865a556833edcda3529fe31406deead3d398

jkbelcher:
Yes, very cool!  This really covers all the bases.  Tested and appears to work great.

PR #43, opened 2022-01-19
APC40mkII - clear lights on disable

jkbelcher:
Clears surface lights including patterns, clips, and device knobs when disabled.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
@@ -318,6 +318,8 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
           if (this.knobs[i] != null) {
             this.knobs[i].removeListener(this);
             this.knobs[i] = null;
+            sendControlChange(0, DEVICE_KNOB + i, 0);
+            sendControlChange(0, DEVICE_KNOB_STYLE + i, LED_STYLE_OFF);
           }
         }
         this.device.controlSurfaceSemaphore.decrement();
@@ -568,6 +570,13 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
     }
   }
 
+  private void clearChannelGrid() {
+    for (int i = 0; i < NUM_CHANNELS; ++i) {
+      sendChannelPatterns(i, null);
+      sendChannelClips(i, null);
+    }
+  }
+
   private void sendChannel(int index, LXAbstractChannel channel) {
     if (channel != null) {
       sendNoteOn(index, CHANNEL_ACTIVE, channel.enabled.isOn() ? LED_ON : LED_OFF);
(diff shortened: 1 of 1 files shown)


mcslee:
Shoot, actually, I am not sure this was a good idea.

unregister() is called AFTER the APC mode has been set back to GENERIC.
https://github.com/heronarts/LX/blob/dev/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java#L499

The APC40mk2 has multiple midi modes. In Ableton control surface mode, the LEDs are fully controlled by MIDI. The device doesn't track any internal state.

In GENERIC mode (default), the device tracks internal state and updates the knobs etc. itself.

With this change, onEnable(false) is going to set the device back to GENERIC mode, and then it's going to send MIDI output messages changing LED things. At this point, after we've turned the device back to GENERIC, we shouldn't be managing its state anymore...

We could do this unregister() operation BEFORE sending the sysex, but TBH I think it's not really necessary. I think once the surface is set back to GENERIC it should be left alone. Apologies for not seeing this before, I think it's probably better to revert this and stick to just relying upon the Sysex message changing APC modes.

jkbelcher:
Setting the APC to GENERIC mode with the sysex resets the focus lights but not the grid lights and not the device control knobs.

[Prior to this update] Oddly, on program exit the pattern grid IS cleared, though not the clip lights if in clip mode and not the device control lights.  Maybe on application exit the pattern lights are getting cleared by an order of operations that disposes the channels before the midi surface.

mcslee:
Ah, okay, this makes sense. On program close, enabled probably never gets turned off, but dispose() will get called which will trigger the unregister() flow.

I think the crucial difference is probably that the unregister() flow in this case happens *without* the device being set back to GENERIC mode.

So maybe one change we want to make in onEnabled(), is that if enabled is true, set to ABLETON_ALTERNATE_MODE immediately, then do the rest of the work.

If enabled is false, do all the unregistration first, and then *after* that set the device back to GENERIC_MODE.

Maybe one to play with? I don't actually have an APC40mk2 here to test with at the moment, so a bit of guesswork.

jkbelcher:
Yes that improves it to unregister() before setting back to GENERIC mode.  Keeps the device control knobs functional in GENERIC mode but sets them back to zero.  Looks clean now.  With my fix the device knobs ended up lights off in GENERIC mode.

Found another stray puzzle piece:
https://github.com/heronarts/LX/blob/dev/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java#L701

This mixer listener wasn't getting removed in unregister().  So it was firing from here:
https://github.com/heronarts/LX/blob/dev/src/main/java/heronarts/lx/mixer/LXMixerEngine.java#L1021
on the LXEngine.dispose() after the midi surfaces are supposed to be disposed.

I'll clean this up too.

mcslee:
Ah nice find, yeah def want to unregister that too. Thanks for chasing this all down!

PR #44, opened 2022-01-20
Clean up shutdown and behavior on disable

jkbelcher:
Set APC GENERIC mode after unregister().  Removes stranded lxmixerengine listener.  Also clear the channel button lights when clearing grid.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
@@ -496,8 +496,8 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
 
   @Override
   protected void onEnable(boolean on) {
-    setApcMode(on ? ABLETON_ALTERNATE_MODE : GENERIC_MODE);
     if (on) {
+      setApcMode(ABLETON_ALTERNATE_MODE);
       initialize(false);
       register();
     } else {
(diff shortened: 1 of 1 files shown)


src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -698,26 +717,8 @@ private void register() {
...
-      public void channelMoved(LXMixerEngine mixer, LXAbstractChannel channel) {
-        sendChannels();
-      }
-
-      @Override
-      public void channelAdded(LXMixerEngine mixer, LXAbstractChannel channel) {
-        sendChannels();
-        registerChannel(channel);
-      }
-    });
```
mcslee, line 720: If you can just toss a this.mixerEngineListener there - I know it's not necessary but I make a habit of doing this. It makes it very clear to the reader that the variable is a member of this class... not of a containing outer class, or a function parameter, etc.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -734,6 +735,7 @@ private void unregister() {
       unregisterChannel(channel);
     }
 
+    this.lx.engine.mixer.removeListener(mixerEngineListener);
```
mcslee: this.

mcslee (changes requested):
Looks good - thanks for cleaning up. Please just throw a "this." prefix on the references to the variable for consistency with LX style. Will merge after that!

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
jkbelcher: Oops.  Got it.

PR #45, opened 2022-01-25
APC40MkII bug fix and clip handling

jkbelcher:
Fixed bug: surface failed to register listeners to existing clips on enable or load from file.

For single property change on a clip: only refresh associated button not entire channel.  It's possible there's a cleaner way to do this... it feels slightly odd due to needing both the clip index and channel index to locate the button.


mcslee:
Nice, thx for that - cleaned up one of my lazy TODOs!

PR #47, opened 2022-02-02
APCmini midi controller

jkbelcher:
Added support for clips and device control.  The button above Shift toggles Parameter mode and the next button up toggles Clip mode.

In parameter/device control mode each parameter gets three buttons - increment, decrement, and reset.  The reset button turns red when a parameter value is non-default; this makes it easy to scan the controller in a "which parameter is out of place" scenario.  For this function I added a boolean isDefault() method to LXListenableParameter.

### DIFF
--- a/src/main/java/heronarts/lx/parameter/LXListenableParameter.java
+++ b/src/main/java/heronarts/lx/parameter/LXListenableParameter.java
@@ -222,6 +222,10 @@ public abstract class LXListenableParameter implements LXParameter {
     return (float) getValue();
   }
 
+  public final boolean isDefault() {
+    return this.getValue() == this.defaultValue;
+  }
+
   public String getLabel() {
     return this.label;
   }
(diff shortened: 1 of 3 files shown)


src/main/java/heronarts/lx/parameter/LXListenableParameter.java
```diff
@@ -222,6 +222,10 @@ public final float getValuef() {
     return (float) getValue();
   }
 
+  public final boolean isDefault() {
+    return this.getValue() == this.defaultValue;
```
mcslee, line 226: FYI, in general LX-style I don't prefix "this." on method calls. Since in Java, it's *always* a method call unless prefixed by StaticClassName.staticFunction().

No big deal to have it in there though - but just spelling out the rational for my convention, why I do "this." on member variables and not on method calls.

mcslee (approved):
FYI didn't look in super close detail at the APCmini.java file, I trust your judgment on that and it seems largely adapted from pre-existing code. But let me know if there were any parts you did specifically want some eyes on.

src/main/java/heronarts/lx/parameter/LXListenableParameter.java
jkbelcher: Got it, thanks.  Will try to keep these aligned.

jkbelcher:
I think it's fine to run without a detailed review.  There are some variations from the other controller but nothing radical.  I'll run it some in the lab and keep an eye out for quirks.

PR #50, opened 2022-02-17
Midi surfaces accommodate null remote controls from device

jkbelcher:
Update midi surfaces to handle null entries in the list of parameters returned by LXDeviceComponent.getRemoteControls().  Devices can now use null as a placeholder when aligning parameters with physical knobs on midi surfaces.  This eliminates the need to use "dummy" placeholder parameters.

Tested and working w/ APC40mkII, APCmini, and MFT.


mcslee:
Nice!

PR #51, opened 2022-02-27
Add palette editing to APC40Mk2

raldi:
I tried recording a video of this code in action, but despite all white-balance-adjustment efforts, I couldn't get the APC's button lights to look onscreen like anything other than uniform shades of white. Instead, here are instructions for making your own walking tour:

1. Open up a project and give it an interesting main palette and a couple extra swatches on the side
2. Press DEV LOCK. Your side swatches should now appear in the channel grid, and the main palette swatch on the SCENE LAUNCH column to the right.
3. Turn CUE LEVEL both with and without SHIFT and note how the APC's button changes color in sync with the onscreen copy.
4. Tap the third (or whatever) row of the SCENE LAUNCH column. This marks the corresponding palette entry as the new focusColor.
5. Repeat Step 3 and note that the knob is now controlling the focusColor.
6. Tap the illuminated Channel Focus button beneath one of your side swatch columns in the grid. That should set this swatch as the main one.
7. Tap one of the illuminated rows in the grid. This copies its color to the clipboard.
8. Tap one of the rows in the SCENE LAUNCH column. This pastes the color from the clipboard. (It also still marks it as the focusColor, so you can tweak it with the CUE LEVEL knob.)
9. Tap another one. It'll paste again.
10. Tap STOP ALL CLIPS. This clears the clipboard, so now if you tap a color in the main palette, it'll only give it focus.
11. Turn the TEMPO knob one notch. The whole grid should light up in a rainbow of colors.
12. Turn the knob some more. The rainbow should scroll.
13. Tap any of these rainbow colors and it'll be copied to the clipboard. As before, tap a row in the SCENE LAUNCH column to paste it and turn the CUE LEVEL knob to tweak.
14. Tap the TAP TEMPO button to turn off Rainbow Mode.
15. Press DEV LOCK to get out of palette mode. (Or BANK, to switch from palette mode to bank mode.)

The most glaring problem remaining is that the RGB values in APC's spec seem kind of bonkers. It may very well be the values being sent to their LEDs, but due to issues of gamma correction or what-have-you, the actual observed result of comparing the illuminated buttons to onscreen representations of their RGB values often shows wild discrepancies. I might have to just go through all 128 colors and rewrite their supposed RGB values to match what my own eye is seeing.


raldi:
I'm redesigning the rainbow picker. The old one was based on the built-in colors of the APC, but who cares about that? The new one will let you scroll through the HSB colorspace, with 72 hues of columns you can scroll through with the Tempo knob, and for each hue, there are five sat/bri combos. The APC will display them with as much fidelity as the hardware supports, and then when you pick one, the color on the clipboard will be the precise HSB for that square, regardless of the APC's limited palette.

raldi:
Also working on rebasing this off the dev branch. If you have the time, though, I'd appreciate a UX / design review of the PR as-is, and I'll integrate your feedback into the next revision.

raldi:
Rebased off the dev branch and redesigned the color picker.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -136,19 +143,32 @@
...
   public static final int LED_MODE_PRIMARY = 0;
   public static final int LED_MODE_PULSE = 10;
   public static final int LED_MODE_BLINK = 15;
 
-  private boolean shiftOn = false;
-  private boolean bankOn = true;
+  // We use three modifier keys:
+  private boolean shiftOn = false;  // All-purpose, momentary, can't light up
+  private boolean bankOn = true;    // Lit toggle. When on, the grid controls patterns instead of clips.
+  private boolean deviceLockOn;     // Lit toggle. When on, much of the hardware is repurposed for palette editing.
```
mcslee: I would explicitly initialize this to false here for parity with the others. Minor thing, but for consistency with the LX codebase in general would put these comments on a single-line before the variable declaration rather than after the var.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -136,19 +143,32 @@
...
+
+  private Integer colorClipboard;   // "Copies" a color for pasting into the main swatch
+  private LXDynamicColor focusColor;// The entry in the main swatch that CUE LEVEL adjusts
+  private boolean rainbowMode;      // Display full spectrum of colors for easy copying
```
mcslee: Again, would initialize this explicitly as with the other fields

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -573,11 +604,16 @@ private void sendChannels() {
...
+    if (!this.rainbowMode) {
+      for (int i = 0; i < NUM_CHANNELS; ++i) {
+        LXAbstractChannel channel = getChannel(i);
+        sendChannelPatterns(i, channel);
+        sendChannelClips(i, channel);
+        sendSwatch(i);
+      }
     }
+
+    sendSwatch(-1);
```
mcslee: I would throw up top a:
private static final int MASTER_SWATCH = -1;

Will make it a lot more clear to the reader what's going on here, it's pretty confusing otherwise.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -622,18 +659,18 @@ private void sendChannelPatterns(int index, LXAbstractChannel channelBus) {
         int color = LED_OFF;
         if (y == activeIndex) {
           // This pattern is active (may also be focused)
-          color = 60;
+          color = LED_ORANGE_RED;
```
mcslee, line 702: Thanks for cleaning these up!

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -603,9 +639,10 @@ private void sendChannel(int index, LXAbstractChannel channel) {
   }
 
   private void sendChannelPatterns(int index, LXAbstractChannel channelBus) {
-    if (index >= CLIP_LAUNCH_COLUMNS || !this.bankOn) {
+    if (index >= CLIP_LAUNCH_COLUMNS || !this.bankOn || this.deviceLockOn) {
```
mcslee: Feels like we might want to introduce a helper method here or an enum that tracks what mode the grid is in. This was already a little bit non-self-explanatory before, when it was directly checking this.bankOn... now I fear it's getting really confusing to an unfamiliar reader.

```
public enum GridMode {
  PATTERN,
  CLIP,
  PALETTE;
}

private GridMode getGridMode() {
  if (this.deviceLockOn) {
    return GridMode.PALETTE;
  } else if (this.bankOn) {
    return GridMode.PATTERN;
  }
  return GridMode.CLIP;
}
```

I think this will save a lot of confusion.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -651,6 +688,9 @@ private void sendChannelPatterns(int index, LXAbstractChannel channelBus) {
 
   private void sendChannelClips(int index, LXAbstractChannel channel) {
     for (int i = 0; i < CLIP_LAUNCH_ROWS; ++i) {
+      int color = LED_OFF;
```
mcslee: Did these three lines get copy-pasted in here? I'm guessing you started a slight refactor but then undid it? I don't see them doing anything now, nuke these lines back out...

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -660,7 +700,8 @@ private void sendChannelClips(int index, LXAbstractChannel channel) {
   }
 
   private void sendClip(int channelIndex, LXAbstractChannel channel, int clipIndex, LXClip clip) {
-    if (this.bankOn || channelIndex >= CLIP_LAUNCH_COLUMNS || clipIndex >= CLIP_LAUNCH_ROWS) {
+    if (this.bankOn || this.deviceLockOn ||
```
mcslee: Ditto here about a GridMode helper

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -678,11 +719,74 @@ private void sendClip(int channelIndex, LXAbstractChannel channel, int clipIndex
...
+      if (swatch != null && i < swatch.colors.size()) {
+        int palColor = swatch.colors.get(i).getColor();
+        color = this.apc40Mk2Colors.nearest(palColor);
+      }
+      sendNoteOn(mode, pitch, color);
+    }
+  }
+
+  private int rainbowGridColor(int hue, int row) {
+    final int[] sats = { 100,  70,  50, 100, 100 };
```
mcslee: If the idea is to only allocate these once (which makes sense), push them outside the function and make them private final static int[]

```
private final static int RAINBOW_GRID_SAT = {... };
private final static int RAINBOW_GRID_BRI = {... };
```

And I think you could even consider just pre-generating the full 2-D array, or lazy-initializing this the first time its asked for. Not a huge deal either way but don't need to be calling LX.hsb() loads of times every time the rainbow is sent... once will do. And in fact you could even just store the result of the nearest color lookup, since that's all you really need.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -678,11 +719,74 @@ private void sendClip(int channelIndex, LXAbstractChannel channel, int clipIndex
...
+    final int[] bris = { 100, 100, 100,  50,  25 };
+    return LXColor.hsb(hue, sats[row], bris[row]);
+  }
+
+  // Cover the grid with a rainbox of APC colors
+  private void sendRainbows() {
```
mcslee: I'd give this a slightly clearer name, like sendGridRainbowPicker(). sendRainbows() is pretty feel-good but feels slightly too vague.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -678,11 +719,74 @@ private void sendClip(int channelIndex, LXAbstractChannel channel, int clipIndex
...
+    int hue = this.rainbowOffset;
+
+    for (int channel = 0; channel < NUM_CHANNELS; ++channel) {
+      for (int row = 0; row < PALETTE_SWATCH_ROWS; ++row) {
+        int pitch = CLIP_LAUNCH + channel + CLIP_LAUNCH_COLUMNS * (CLIP_LAUNCH_ROWS - 1 - row);
+        int color = rainbowGridColor(hue, row);
+        int colorId = apc40Mk2Colors.nearest(color);
```
mcslee, line 814: Per above, consider pre-generating the full array of nearest color ids, skip all this dynamic color generation+lookup everytime the rainbow picker is sent.

src/main/java/heronarts/lx/color/LXColor.java
```diff
@@ -53,6 +53,8 @@
   public static final int R_SHIFT = 16;
   public static final int G_SHIFT = 8;
 
+  public static final double MAX_HUE = 360;
```
mcslee: Fine to make this an int, which is better since you're doing int modular arithmetic with it in the other file.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -678,11 +719,74 @@ private void sendClip(int channelIndex, LXAbstractChannel channel, int clipIndex
...
+        sendNoteOn(LED_MODE_PRIMARY, pitch, colorId);
+      }
+      hue += RAINBOW_HUE_STEP;
```
mcslee: Think it's better to write these on one line, as it's really one update operation and that'll defend against a bug where one line is accidentally changed without the other.

`hue = (hue + RAINBOW_HUE_STEP) % LXColor.MAX_HUE;`

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -792,17 +906,43 @@ private void noteReceived(MidiNote note, boolean on) {
     case BANK:
       if (on) {
         this.bankOn = !this.bankOn;
+        if (this.deviceLockOn) {
```
mcslee, line 953: I would maybe have a helper function here that performs the operations when the grid mode changes. These three lines are very related to the lines below for handling when DEVICE_LOCK is pressed. Feels like this could be more clear, with less conditional nesting in one helper.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -858,18 +997,54 @@ private void noteReceived(MidiNote note, boolean on) {
         this.lx.engine.mixer.cueB.toggle();
         return;
       case STOP_ALL_CLIPS:
-        this.lx.engine.mixer.stopClips();
+        if (this.deviceLockOn) {
+          this.colorClipboard = null;
```
mcslee, line 1046: Again, sort of seems fine to this either not even be here, or just set to 0 - as far as I can tell clearing the clipboard has no real side effect.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -858,18 +997,54 @@ private void noteReceived(MidiNote note, boolean on) {
...
       }
 
       if (pitch >= SCENE_LAUNCH && pitch <= SCENE_LAUNCH_MAX) {
-        this.lx.engine.mixer.launchScene(pitch - SCENE_LAUNCH);
+        int index = pitch - SCENE_LAUNCH;
+        if (this.deviceLockOn) {
+          // A button corresponding to an entry in the main palette swatch has been tapped.
+          // Make it the focusColor (i.e., turning the CUE LEVEL knob will tweak it), and
+          // if there's a color on the clipboard, paste it here.
+          LXSwatch swatch = getSwatch(-1);
```
mcslee: `MASTER_SWATCH`

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -901,8 +1076,31 @@ private void noteReceived(MidiNote note, boolean on) {
...
 
+
     // Channel messages
+
+    if (this.rainbowMode) {
+      return;
+    }
+
+    if (this.deviceLockOn) {
+      if (!on) switch (note.getPitch()) {
```
mcslee: Oof, clever line-saver but that feels scary to me... mind breaking this out to a fully braced thing?

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -941,14 +1139,10 @@ private void noteReceived(MidiNote note, boolean on) {
...
       case DEVICE_LEFT:
-        this.deviceListener.registerPrevious();
-        return;
-      case DEVICE_RIGHT:
-        this.deviceListener.registerNext();
-        return;
       case BANK_LEFT:
         this.deviceListener.registerPrevious();
         return;
+      case DEVICE_RIGHT:
```
mcslee, line 1188: Nice, thx again

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -973,18 +1167,29 @@ public void controlChangeReceived(MidiControlChange cc) {
     int number = cc.getCC();
     switch (number) {
     case TEMPO:
-      if (this.shiftOn) {
+      if (this.deviceLockOn) {
+        if (this.rainbowMode) {
+          this.rainbowOffset += RAINBOW_HUE_STEP * cc.getRelative();
```
mcslee: Again would combine this all into a single expression:
`this.rainbowOffset = (this.rainbowOffset + LXColor.MAX_HUE + RAINBOW_HUE_STEP * cc.getRelative()) % LXColor.MAX_HUE;`

I think it's actually clearer what's happening in one line than 3.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -973,18 +1167,29 @@ public void controlChangeReceived(MidiControlChange cc) {
...
       return;
     case CUE_LEVEL:
       if (this.shiftOn) {
-        this.lx.engine.palette.color.primary.saturation.incrementValue(cc.getRelative());
+        this.focusColor.primary.saturation.incrementValue(cc.getRelative());
       } else {
-        this.lx.engine.palette.color.primary.hue.incrementValue(cc.getRelative(), true);
+        this.focusColor.primary.hue.incrementValue(cc.getRelative(), true);
       }
+      if (this.deviceLockOn) sendSwatch(-1);
```
mcslee: LX style nit-pick, throw some braces around there just for good measure and consistency

mcslee (changes requested):
Looks pretty solid overall. No major problems with this, just a list of relatively minor refactoring and style cleanup stuff. I think the bankOn / deviceLockOn code is getting too opaque now that there are 3 (and a half) modes the grid can be in. Let's make that more clear by spelling it out with some helper functions and probably an enum that semantically identifies the different modes. I think the code will become a lot more clear this way, as it will become readable what mode the grid is in, as well as the relationship between the buttons and the modes.

The only other high-level thing I'll throw out here - the control surface right now does not do any listening to the palette or color swatches. If the user makes modifications to the palette colors or swatch while the device lock is on, the control surface will not update to reflect those changes. I am fine with this, as it keeps the code a lot simpler and I think this is acceptable given the intended workflow looking like the user pops temporarily into palette mode by hitting dev lock, works on the surface, and then pops out again.

But just throwing it out there - a more complete implementation would add a bunch of listeners to the palette and the swatch colors to catch any modifications to those, and would update the control surface accordingly if it is in the palette mode at the time. Again, I do actually recommend holding off on this for now, it's way more surface area for bugs and issues and the added value to the UX is somewhat marginal. As it stands right now, the user can always "fix" the APC by just hitting dev lock twice, so I think that's fine.

Thanks for knocking this out!

mcslee:
Also to clarify my suggestion the GridMode stuff - I think it's probably worth actually making that a member variable of the class, and having the value set + update when either the BANK or the DEVICE_LOCK button is pushed.

That way all of the other code that's checking those things can simply check:
`if (this.gridMode == GridMode.WHICHEVER)`

And there can be:
`private void setGridMode()`

invoked when either BANK or DEV LOCK is pressed, and it can deal with updating `this.gridMode` and performing all of the necessary side-effect operations.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Initializing here introduces the potential for inconsistency with resetPaletteVars(), but I see the value of what you're saying. Done.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Great suggestion. Done.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done. I also inverted the logic to match the others: The action happens on button down, not button up.

src/main/java/heronarts/lx/color/LXColor.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: This was needed when I was based off the latest stable release, but when I rebased to dev, it became vestigial. Removed.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: We need to know the actual color too, so we can set the swatch entry to it, but pregeneration is a good idea. Done.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Right on both counts. Done.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: The nearest-ID table is cached, so it's already just a lookup.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Without pass-by-reference, I don't see any way to do this.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: I'm not sure what you're suggesting here.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

raldi:
I have some big ideas involving dynamic colors, but I'd like to get this PR in first and then I'll write them up for your consideration.

Next iteration of this PR is up.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
mcslee: Sorry I had meant to delete that comment.

mcslee:
Good stuff, just merged and made a few very minor cleanups. Some of that bank-vs-device button stuff could perhaps go into the updateGridMode() method to ensure that they are always consistent, but I think it's fine how you've done it.

https://github.com/heronarts/LX/commit/0fcb6a546ed6a65edeb8836fdc7382a05505afbf

Ping me about dynamic colors whenever you like!

PR #52, opened 2022-02-28
Clip-loop features for the APC40Mk2

raldi:
On the APC40Mk2, you can now hold SHIFT and click a clip to toggle whether it loops. This works whether it's currently playing or not.

When idle, clips that loop show up in a slightly more cyan shade of gray than the one used for non-looping clips. When playing, they pulse between green and cyan rather than green and off.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
@@ -436,6 +436,7 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
       for (LXClip clip : this.channel.clips) {
         if (clip != null) {
           clip.running.addListener(this);
+          clip.loop.addListener(this);
         }
       }
     }
@@ -460,6 +461,7 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
       for (LXClip clip : this.channel.clips) {
         if (clip != null) {
           clip.running.removeListener(this);
+          clip.loop.removeListener(this);
         }
       }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/heronarts/lx/mixer/LXBus.java
```diff
@@ -333,10 +333,14 @@ public LXClip getClip(int index, boolean create) {
...
   public LXClip addClip() {
-    return addClip(this.mutableClips.size());
+    return addClip(this.mutableClips.size(), false);
   }
 
   public LXClip addClip(int index) {
+    return addClip(index, false);
+  }
+
+  public LXClip addClip(int index, boolean loop) {
```
mcslee: This feels strange, I wouldn't start this pattern in general because components have loads of parameters. These parameters should be changed post-core API method and listeners should do the right side effects. Let's leave this file untouched.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -1105,14 +1108,15 @@ private void noteReceived(MidiNote note, boolean on) {
           } else {
             LXClip clip = channel.getClip(index);
             if (clip == null) {
-              clip = channel.addClip(index);
+              channel.addClip(index, this.shiftOn);
```
mcslee: Change this to `clip = channel.addClip(index)` followed by:
```
if (this.shiftOn) {
  clip.loop.setValue(true);
}
```

Then pop over to lines 548 - 559, and add:
```
clip.loop.addListener(this);
clip.loop.removeListener(this);
```

Adjacent to where clip.running is monitored. This will ensure the right thing happens whenever and however the looping status of the clip is modified, whether by control surface or otherwise.

Also need to do that on lines 437/461 in the ChannelListener contructor and dispose method.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -1105,14 +1108,15 @@ private void noteReceived(MidiNote note, boolean on) {
...
+            } else if (this.shiftOn) {
+              clip.loop.toggle();
+              sendClip(channelIndex, channel, index, clip);
```
mcslee: You'll be able to drop this line when we add the listener to clip.loop - the parameter listener will get the clip redrawn properly.

mcslee (changes requested):
Nice, I dig the idea. Let's leave LXBus untouched and add a listener to clip.loop in the APC40mk2 so that it handles loop changes properly wherever they originate.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done, though I inlined line 1116 to avoid another layer of braces.

PR #53, opened 2022-03-02
WIP: Add hooks so apps can grab colors from APC

raldi:
Added a hook into the APC40Mk2 driver so that app code can peek in and see if there's a color that was recently touched, and if so, whether it's a fixed value or a swatch entry. The changes to SolidPattern aren't actually intended to be checked in; they're just my attempt to show how an app could hook into this.

### DIFF
--- a/src/main/java/heronarts/lx/LXComponent.java
+++ b/src/main/java/heronarts/lx/LXComponent.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import com.google.gson.JsonObject;
 import heronarts.lx.color.ColorParameter;
 import heronarts.lx.color.DiscreteColorParameter;
+import heronarts.lx.color.LinkableColorParameter;
 import heronarts.lx.modulation.LXModulationContainer;
 import heronarts.lx.osc.LXOscComponent;
 import heronarts.lx.osc.LXOscEngine;
@@ -884,10 +885,18 @@ public abstract class LXComponent implements LXPath, LXParameterListener, LXSeri
     }
     if (parameter instanceof ColorParameter) {
       // NOTE: order is important, brightness must be saved/loaded first
+      // TODO: Why?
       ColorParameter colorParameter = (ColorParameter) parameter;
       addParameter(path + "/" + ColorParameter.PATH_BRIGHTNESS, colorParameter.brightness);
       addParameter(path + "/" + ColorParameter.PATH_SATURATION, colorParameter.saturation);
       addParameter(path + "/" + ColorParameter.PATH_HUE, colorParameter.hue);
+    }
+    if (parameter instanceof LinkableColorParameter) {
+      LinkableColorParameter linkableColorParameter = (LinkableColorParameter) parameter;
+      addParameter(path + "/" + LinkableColorParameter.PATH_SWATCH_NUMBER,
+              linkableColorParameter.swatchNumber);
+      addParameter(path + "/" + LinkableColorParameter.PATH_SWATCH_COLOR_NUMBER,
+              linkableColorParameter.swatchColorNumber);
     }
     return this;
   }
--- a/src/main/java/heronarts/lx/color/ColorParameter.java
+++ b/src/main/java/heronarts/lx/color/ColorParameter.java
@@ -33,7 +33,7 @@ public class ColorParameter extends LXListenableParameter implements LXParameter
   public static final String PATH_SATURATION = "saturation";
   public static final String PATH_BRIGHTNESS = "brightness";
 
-  private int color;
+  protected int color;
   private boolean internalValueUpdate = false;
   private boolean internalHsbUpdate = false;
 
@@ -53,6 +53,9 @@ public class ColorParameter extends LXListenableParameter implements LXParameter
     this.hue.addListener(this);
     this.saturation.addListener(this);
     this.brightness.addListener(this);
+    this.hue.parentParameter = this;
+    this.saturation.parentParameter = this;
+    this.brightness.parentParameter = this;
     this.color = color;
   }
 
@@ -71,7 +74,7 @@ public class ColorParameter extends LXListenableParameter implements LXParameter
   }
 
   public String getHexString() {
-    return String.format("0x%08x", this.color);
+    return String.format("0x%08x", this.getColor());
   }
 
   @Override
--- a/src/main/java/heronarts/lx/LX.java
+++ b/src/main/java/heronarts/lx/LX.java
@@ -31,7 +31,7 @@ import heronarts.lx.output.LXOutput;
 import heronarts.lx.parameter.MutableParameter;
 import heronarts.lx.parameter.StringParameter;
 import heronarts.lx.pattern.LXPattern;
-import heronarts.lx.pattern.color.SolidPattern;
+import heronarts.lx.pattern.color.NewSolidPattern;
 import heronarts.lx.scheduler.LXScheduler;
 import heronarts.lx.structure.LXFixture;
 import heronarts.lx.structure.LXStructure;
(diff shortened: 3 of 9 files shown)


mcslee:
Interesting, I think it's reasonable to make that stuff accessible from the APC class, but feels like a weird structure for application code to reach into a particular control surface implementation's guts. I think it's better if application code registers or signals "I want this thing to be controllable by surfaces" and then the surface is just able to make the modification.

Certainly wouldn't want to get into a situation where pattern/app code had to make N checks for N different types of control surface.

As an illustration of saving the color state though, this makes sense.

I'm unfortunately super-backlogged on getting much work done at the moment as we've all got COVID over here so no nanny at the moment... but hopefully get through it this week and I can push forward on this AggregateParameter/ColorParameter stuff next week.

raldi:
In the meantime, here's a much-improved proof-of-concept. To give yourself a tour:

1. Create a few nontrivial swatches and a channel running a NewSolidPattern
2. Note that Device Control knob 1 is the only one in use. This single knob is all you need to control all aspects of the color: hue, saturation, brightness, linked swatch number, linked swatch color index!
3. Turn it to adjust the hue. (Note that it wraps around)
4. Press "<-- Bank (3)"; it'll light up. Now turning the knob controls saturation.
5. Press "Bank --> (4)"; this will light up and the other will go dark. Now turning the knob controls brightness.
6. Press "Bank --> (4)" again so that both are dark and the knob is back to controlling hue. Hold shift while turning it, and you control saturation (for backwards compatibility with existing muscle memory)
7. Press "Dev. Lock (6)" to turn on swatch mode.
8. Tap a grid button corresponding to a swatch color. It's now "the focus".
9. Turn the knob any amount. It is now linked to the focused swatch color.
10. Turn CUE LEVEL. It'll adjust the swatch color (honoring SHIFT, "BANK ->" and "<- BANK" just like the device knob did) and thus, the linked pattern color.
11. Press STOP ALL CLIPS to clear the focus.
12. Repeat steps 8-10 but with a color from the master swatch.
13. Turn the TEMPO knob to activate the picker.
14. Tap anywhere in the grid to pick a color. It's now the focused fixed color.
15. Turn the device knob. It'll now be set to the focused fixed color.

raldi:
16. Press TAP TEMPO to turn off picker mode
17. Press a grid button to select a swatch color.
18. Turn the device control knob to link the pattern color to the swatch color.
19. Press DEV LOCK to go back to Clip or Pattern mode.
20. Turn the device control knob; it will change which swatch the color is linked to (maintaining the same swatch color index)
21. Hold SHIFT and turn it; it will change which swatch color index it's linked to (maintaining the same linked swatch)

src/main/java/heronarts/lx/LXComponent.java
```diff
@@ -884,11 +885,19 @@ protected LXComponent addParameter(String path, LXParameter parameter) {
     }
     if (parameter instanceof ColorParameter) {
       // NOTE: order is important, brightness must be saved/loaded first
+      // TODO: Why?
```
mcslee, line 888: When any of the HSB params changes, the color is updated, and vice versa. But there's potentially more information in the HSB params than the 32-bit color. Particularly, e.g. if brightness or saturation is 0, then there's no hue information to extract from the 32-bit color value. The position of the color slider is lost.

So HUE is restored LAST to stop the later restoration of a potentially-0 brightness/saturation value from clobbering it.

I am not sure this is *strictly* necessary anymore, I may have fixed it another way (and I aim to make this cleaner with AggregateParameter).

src/main/java/heronarts/lx/LXComponent.java
```diff
@@ -884,11 +885,19 @@ protected LXComponent addParameter(String path, LXParameter parameter) {
...
       ColorParameter colorParameter = (ColorParameter) parameter;
       addParameter(path + "/" + ColorParameter.PATH_BRIGHTNESS, colorParameter.brightness);
       addParameter(path + "/" + ColorParameter.PATH_SATURATION, colorParameter.saturation);
       addParameter(path + "/" + ColorParameter.PATH_HUE, colorParameter.hue);
     }
+    if (parameter instanceof LinkableColorParameter) {
```
mcslee, line 894: This is the kind of special-casing mess I want to clean up with AggregateParameter... more to come on that soon when I find time...

src/main/java/heronarts/lx/LXComponent.java
```diff
@@ -884,11 +885,19 @@ protected LXComponent addParameter(String path, LXParameter parameter) {
...
+      LinkableColorParameter linkableColorParameter = (LinkableColorParameter) parameter;
+      addParameter(path + "/" + LinkableColorParameter.PATH_SWATCH_NUMBER,
```
mcslee, line 896: I don't love referencing swatch number, this will break in weird ways if a swatch is deleted, or swatches are reordered. I'm not really sure that it should be possible to reference saved swatches vs. referencing the active swatch.

I had anticipated this desire before, as it particularly seems tempting to want two active swatches in a crossfader situation (e.g. swatch A and B for either side). But it's a potentially slippery slope to active-swatches-upon-swatches... I'll mull this further.

Referencing an active swatch color position by index def feels good though.

src/main/java/heronarts/lx/color/ColorParameter.java
```diff
@@ -53,6 +53,9 @@ public ColorParameter(String label, int color) {
     this.hue.addListener(this);
     this.saturation.addListener(this);
     this.brightness.addListener(this);
+    this.hue.parentParameter = this;
```
mcslee, line 56: AggregateParameter will make this sort of thing a lot cleaner...

mcslee (commented):
Conceptually this looks about right. I'll keep you posted when the AggregateParameter stuff is ready as I think that will clean up some of the parameter management side of this.

I'm not sure on the ability to reference swatches by index... that's going to be brittle, it won't be resilient to modifications to the saved swatch list. Need to think more on that, may just want to start with active-swatch only for color references. Whether to have multiple active swatches has been on my mind for a bit... I've tried to hold off as it complicates various UX things.

Hopefully will have time to dive into this properly soon...

mcslee:
Apologies this took forever for me to get to. I have refactored and worked out AggregateParameter which plays nicely with all the things it needs to, like component saving, snapshots, OSC messages, etc. - there's a decent # of side effects from these sorts of magical parameters.
https://github.com/heronarts/LX/commit/958b3a292862558a75c24cfa7e69db33d0f06771

With that foundation, ColorParameter and LinkedColorParameter should be more manageable - updated those and the SolidColorPattern to use them. Also made the assignment and access of getParentParameter() a bit cleaner, a first-class LXParameter method.

I will say, however, that I don't love that needing to be accessed much in general. I feel a little weird about the APC40 looking inside of the parameters it's passed and doing all this special stuff, but I think the functionality is compelling enough that I'm probably willing to let it slide on this one - so long as the knob still behaves in a "default" way like it's expected to (e.g. as a simple knob rotating a hue value).

I think my changeset provides everything needed here such that you should *only* need to change the APC40 file to implement the behaviors you want here, subject to the limitation that this version of LinkedColorParameter does not allow for referencing saved swatches by index, only the active swatch. I think there is a bit too much brittleness with referencing saved swatches by index, when they are moved/rearranged or mutated with the undo stack the index references will go stale and I think there is too much surface area for confusion there.

Let me know if that's all cool by you - hopefully not too much of a pain to rebase this against the AggregateParameter update. Like I said I think you'll be able to mostly just chuck the changes from outside of APC40.

raldi:
I'll merge in the latest changes and give it a shot, thanks!

raldi:
This work has been superseded by https://github.com/heronarts/LX/pull/58 ; you can close.

PR #54, opened 2022-03-13
Add color to swatch when unused master swatch button clicked in Apc40mkII palette mode

jkbelcher:
Just got to testing this new palette mode - very cool!  Quick mod here to allow a new color to be added to the current swatch by clicking one of the unused Scene Launch buttons.  Initializes to the clipboard color if one has been copied or the default color if not.

If initializing to the default color the button doesn't light up until cue level is adjusted.  Fine for now but will be cleaner once midi surface registers for the palette listeners.

Might be nice to choose a button as well to delete the focused color.  Maybe "Nudge -"?

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
@@ -1067,6 +1067,9 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
           // Make it the focusColor (i.e., turning the CUE LEVEL knob will tweak it), and
           // if there's a color on the clipboard, paste it here.
           LXSwatch swatch = getSwatch(MASTER_SWATCH);
+          if (index > swatch.colors.size()-1 && index < LXSwatch.MAX_COLORS) {
+            swatch.addColor();
+          }
           this.focusColor = swatch.getColor(index);
           if (this.colorClipboard != null) {
             this.focusColor.primary.setColor(this.colorClipboard);
--- a/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
@@ -1066,13 +1066,18 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
           // A button corresponding to an entry in the main palette swatch has been tapped.
           // Make it the focusColor (i.e., turning the CUE LEVEL knob will tweak it), and
           // if there's a color on the clipboard, paste it here.
+          boolean colorChanged = false;
           LXSwatch swatch = getSwatch(MASTER_SWATCH);
           if (index > swatch.colors.size()-1 && index < LXSwatch.MAX_COLORS) {
             swatch.addColor();
+            colorChanged = true;
           }
           this.focusColor = swatch.getColor(index);
           if (this.colorClipboard != null) {
             this.focusColor.primary.setColor(this.colorClipboard);
+            colorChanged = true;
+          }
+          if (colorChanged) {
             sendSwatch(MASTER_SWATCH);
           }
         } else {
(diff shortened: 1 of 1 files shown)


raldi:
I have a couple quick notes; can you add me as a reviewer so I can annotate the PR?

jkbelcher:
Looks like only Mark can add reviewers.  Until then I think you can comment directly on the lines if that works?

raldi:
Looks like I can only comment here, but that's fine. I think this is a good idea for an improvement, but you should make the same change on line 1095 so it works with swatches, too.

Also, just a nitpick, but on line 1070, instead of
index > swatch.colors.size()-1
why not do
index >= swatch.colors.size()
?

jkbelcher:
> Looks like I can only comment here, but that's fine. I think this is a good idea for an improvement, but you should make the same change on line 1095 so it works with swatches, too.
Yes might as well include it there since the UI allows adding a color with right click -> Add Color.  Questions:
1. By default through the engine/ui, adding a color to swatch duplicates the last color in the swatch.  But is there any way to modify that color?  I haven't used the color palette extensively, can't see if there is a way to overwrite an existing swatch after recalling it to master and editing one of the colors.
2. It seems like the way to go here is to mimic the midi controller's master swatch behavior, where a newly added color is initialized to the clipboard if available or to the last color in the swatch if not.  That's extra behavior beyond the UI capability but seems to fit well with the rest of this grid mode behavior.  Does that seem right?

> Also, just a nitpick, but on line 1070, instead of index > swatch.colors.size()-1 why not do index >= swatch.colors.size() ?

Good question.  This is a stylistic element on my part.  Approaching it like a math equation... what I want to know is: "Does the index fall outside of the range of the existing indexes?"  The function is exclusive, and < and > are exclusive operators. 
 I'm looking for this:
   if (index > maxIndex)
...and then I substitute size()-1 for maxIndex to get:
   if (index > thing.size()-1)
When I come along later and read this, I know that a size()-1 or length-1 is a conversion from a 1-based index to a 0-based index.  I don't mind the -1s because that reads as a context shift operation.  Then I've retained some readable information in my operator... because > is exclusive I can see at a glance that this operation is looking for something outside the range of the other things.  Whereas I would more likely use <= and >= as inclusive operators in a clearly inclusive operation like:
while (thing <= MAX_SIZE) { ... }

I wouldn't say I'm religious about this style, just explaining this to show that it is for a reason.  Although you could trade the -1 for an = in this case, I don't see it as an optimization when it's a loss of clarity to be doing an exclusive operation with an inclusive operator.  But more important than this type of style would be consistency across a project, so deference to that as well.

If I was strict about always comparing like contexts I might have done:
   if (index > swatch.colors.size()-1 && swatch.colors.size() < LXSwatch.MAX_COLORS) { }
...but I lean away from that one since I already have a relevant integer in a local variable, rather than do a method call that could potentially trigger a more expensive calculation.

And of course I break from this style as well and do for loops like everyone else:
   for (int i=0; i<thing.size(); i++) { }
...in this case removing two operators is worth the visual cleanup from what would otherwise be:
   for (int i=0; i<=thing.size()-1; i++) { }

In summary, had a reason, but I can think of plenty of variations I make to my own style.

raldi:
> By default through the engine/ui, adding a color to swatch duplicates the last color in the swatch. But is there any way to modify that color? 

Sure, just click it and you get a picker.

> It seems like the way to go here is to mimic the midi controller's master swatch behavior, where a newly added color is initialized to the clipboard if available or to the last color in the swatch if not. 

Sounds reasonable.

As for the stylistic choice, I'm fine either way.

jkbelcher:
> Sure, just click it and you get a picker.

Ah, thanks.

Made a couple changes - how does that look?  Added a refresh for the scene launch buttons too, they might as well behave right.

raldi:
This looks great!

mcslee:
Dunno why you couldn't do a review... I just added you both as repository collaborators, maybe this will help for future pull requests and enable formal reviews? In any case looks like you guys got this sorted, seems sensible to me, merging!

PR #55, opened 2022-03-13
Use relative knob changes on MFT and add incrementNormalized() to parameters

jkbelcher:
The MFT can be configured to send increase/decrease MIDI CCs on knob rotation instead of absolute (0-127) position.  I think this is a preferable setting for use with LX.  It avoids an annoying MFT firmware limitation where the hardware device always sets the encoder level indicators when using absolute position.  This makes it impossible to see on the MFT when a discreteParameter has incremented to a new value.  With the knobs sending relative CCs ("ENC 3FH/41H" midi type) the indicator levels are set only by returning MIDI which retains more control from LX.

This also opens up the possibility of wrappable parameters which is super useful for things like hue or degrees of rotation.  A parameter can be made wrappable like this:
public final CompoundParameter angle = 
    (CompoundParameter) new CompoundParameter("Angle", .5, 0, 1)
    .setWrap(true);

I added some methods for wrap and incrementNormalized() for use by the MFT but not sure if LXListenableNormalizedParameter is the right home for them.  They're mainly for BoundedParameter and DiscreteParameter which are some but not all of the subclasses of LXListenableNormalizedParameter.

DiscreteParameters are now incremented/decremented by the MFT with a fixed amount of knob rotation regardless of the number of items in the list.  From my perspective this is a preferred behavior; it felt inconsistent when Enum(30 items) and Enum(4 items) required different amounts of rotation to increment one item a list.  However if someone is using DiscreteParameters with 100+ values this might not be the desired behavior.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
+++ b/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
@@ -39,6 +39,7 @@ import heronarts.lx.parameter.DiscreteParameter;
 import heronarts.lx.parameter.LXListenableNormalizedParameter;
 import heronarts.lx.parameter.LXParameter;
 import heronarts.lx.parameter.LXParameterListener;
+import heronarts.lx.parameter.DiscreteParameter.IncrementMode;
 import heronarts.lx.pattern.LXPattern;
 import heronarts.lx.utils.LXUtils;
 
@@ -192,11 +193,13 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
     private final LXListenableNormalizedParameter[] knobs =
       new LXListenableNormalizedParameter[DEVICE_KNOB_NUM];
     private final int[] knobTicks = new int[DEVICE_KNOB_NUM];
+    private final int[] knobIncrementSize = new int[DEVICE_KNOB_NUM];
 
     DeviceListener() {
       for (int i = 0; i < this.knobs.length; ++i) {
         this.knobs[i] = null;
         this.knobTicks[i] = 0;
+        this.knobIncrementSize[i] = 1;
       }
     }
 
@@ -280,6 +283,11 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
             this.knobs[i] = parameter;
             this.knobTicks[i] = 0;
             if (parameter != null) {
+              // We will track an absolute knob value for Normalized DiscreteParameters even though MFT sends relative CCs.
+              if (parameter instanceof DiscreteParameter && ((DiscreteParameter)parameter).getIncrementMode() == IncrementMode.NORMALIZED) {
+                this.knobTicks[i] = (int) (parameter.getNormalized() * 127);
+                this.knobIncrementSize[i] = LXUtils.max(1, 128/((DiscreteParameter)parameter).getRange());
+              }
               parameter.addListener(this);
               sendControlChange(CHANNEL_ANIMATIONS_AND_BRIGHTNESS, DEVICE_KNOB + i, INDICATOR_ANIMATION_NONE);
               sendControlChange(CHANNEL_ANIMATIONS_AND_BRIGHTNESS, DEVICE_KNOB + i, INDICATOR_BRIGHTNESS_MAX);
@@ -320,7 +328,6 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
           sendControlChange(CHANNEL_ANIMATIONS_AND_BRIGHTNESS, DEVICE_KNOB + i, INDICATOR_BRIGHTNESS_25);
           sendControlChange(CHANNEL_ROTARY_ENCODER, DEVICE_KNOB+i, 0);
           sendControlChange(CHANNEL_ANIMATIONS_AND_BRIGHTNESS, DEVICE_KNOB + i, RGB_BRIGHTNESS_OFF);
-          this.knobTicks[i] = 0;
           ++i;
         }
 
@@ -343,6 +350,11 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
             double normalized = (parameter instanceof CompoundParameter) ?
               ((CompoundParameter) parameter).getBaseNormalized() :
               this.knobs[i].getNormalized();
+            // Normalized DiscreteParameters need artificial tracking of absolute knob location.
+            // Keep local tracking in sync with changes from other source.
+            if (parameter instanceof DiscreteParameter && ((DiscreteParameter)parameter).getIncrementMode() == IncrementMode.NORMALIZED) {
+              this.knobTicks[i] = (int) (normalized * 127);
+            }
             sendControlChange(CHANNEL_ROTARY_ENCODER, DEVICE_KNOB + i, (int) (normalized * 127));
             break;
           }
@@ -367,18 +379,35 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
       LXListenableNormalizedParameter knob = this.knobs[index];
       if (knob != null) {
         if (knob instanceof DiscreteParameter) {
-          // Move after a set number of ticks in the same direction
-          if (isUp) {
-            this.knobTicks[index] = LXUtils.max(this.knobTicks[index], 0) + 1;
-          } else {
-            this.knobTicks[index] = LXUtils.min(this.knobTicks[index], 0) - 1;
-          }
-          if (this.knobTicks[index] == KNOB_TICKS_PER_DISCRETE_INCREMENT * (isUp ? 1 : -1)) {
-            this.knobTicks[index] = 0;
-            if (isUp) {
-              ((DiscreteParameter)knob).increment();
+          if (((DiscreteParameter)knob).getIncrementMode() ==  IncrementMode.NORMALIZED) {
+            int value = this.knobTicks[index] + (isUp ? 1 : -1);
+            if (knob.isWrappable()) {
+              // Make the length of the wrap space the same as the length between other values on this parameter
+              if (value < 0-this.knobIncrementSize[index] || value > 127+this.knobIncrementSize[index]) {
+                value = value < 0 ? 127 : 0;
+              }
+              this.knobTicks[index] = value;
+              value = LXUtils.constrain(value, 0, 127);
             } else {
-              ((DiscreteParameter)knob).decrement();
+              value = LXUtils.constrain(value, 0, 127);
+              this.knobTicks[index] = value;
+            }
+            knob.setNormalized(value/128.);
+          } else {
+            // IncrementMode == RELATIVE
+            // Move after a set number of ticks in the same direction
+            if (isUp) {
+              this.knobTicks[index] = LXUtils.max(this.knobTicks[index], 0) + 1;
+            } else {
+              this.knobTicks[index] = LXUtils.min(this.knobTicks[index], 0) - 1;
+            }
+            if (this.knobTicks[index] == KNOB_TICKS_PER_DISCRETE_INCREMENT * (isUp ? 1 : -1)) {
+              this.knobTicks[index] = 0;
+              if (isUp) {
+                ((DiscreteParameter)knob).increment();
+              } else {
+                ((DiscreteParameter)knob).decrement();
+              }
             }
           }
         } else {
@@ -421,6 +450,8 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
           if (this.knobs[i] != null) {
             this.knobs[i].removeListener(this);
             this.knobs[i] = null;
+            this.knobTicks[i] = 0;
+            this.knobIncrementSize[i] = 1;
           }
         }
         this.device.controlSurfaceSemaphore.decrement();
--- a/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
+++ b/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
@@ -40,6 +40,7 @@ import heronarts.lx.parameter.LXListenableNormalizedParameter;
 import heronarts.lx.parameter.LXParameter;
 import heronarts.lx.parameter.LXParameterListener;
 import heronarts.lx.pattern.LXPattern;
+import heronarts.lx.utils.LXUtils;
 
 public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.Bidirectional {
 
@@ -56,8 +57,9 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
    *     Right Button 2 Function: Next Bank
    *     Right Button 3 Function: CC Toggle
    * Encoder Settings (click Multiple, select all encoders):
+   *   Sensitivity: High Resolution
    *   Switch Action Type: CC Hold
-   *   Encoder MIDI Type: CC
+   *   Encoder MIDI Type: ENC 3FH/41H
    *   Encoder Switch MIDI Settings:
    *     Switch MIDI Channel: 2
    *     Switch MIDI Number: 6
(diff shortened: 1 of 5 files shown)


src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -1067,6 +1067,9 @@ private void noteReceived(MidiNote note, boolean on) {
           // Make it the focusColor (i.e., turning the CUE LEVEL knob will tweak it), and
           // if there's a color on the clipboard, paste it here.
           LXSwatch swatch = getSwatch(MASTER_SWATCH);
+          if (index > swatch.colors.size()-1 && index < LXSwatch.MAX_COLORS) {
```
mcslee: I don't think this part belonged in this pull request?

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
```diff
@@ -337,12 +350,57 @@ public void onParameterChanged(LXParameter parameter) {
       }
     }
 
+    private boolean isKnobRelative(int index) {
+      // Fixed for now but could be expanded to poll MFT encoder settings on startup with sysex
```
mcslee: Is it possible to set the knob mode via MIDI? Or does the MFT user need to know that they need to do some button presses on the MFT to get this configured properly?

Ideally the control surface mode should activate the appropriate mode on all the encoders.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
```diff
@@ -337,12 +350,57 @@ public void onParameterChanged(LXParameter parameter) {
...
 
+    private final static double KNOB_INCREMENT_AMOUNT = 1.0/128.0;
+
+    private void onKnobDecrement(int index) {
+      LXListenableNormalizedParameter knob = this.knobs[index];
+      if (knob != null) {
+        if (knob instanceof DiscreteParameter) {
+          // Move after a set number of ticks in the same direction
+          this.knobTicks[index] = LXUtils.min(this.knobTicks[index], 0) - 1;
+          if (this.knobTicks[index] == 0-KNOB_TICKS_PER_DISCRETE_INCREMENT) {
```
mcslee: `0-KNOB_TICKS_PER_DISCRETE_INCREMENT` feels weird, just put `-KNOB_TICKS_PER_DISCRETE_INCREMENT` without the leading 0

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
```diff
@@ -337,12 +350,57 @@ public void onParameterChanged(LXParameter parameter) {
...
+            this.knobTicks[index] = 0;
+            ((DiscreteParameter)knob).decrement();
+          }
+        } else {
+          onKnobIncrement(knob, 0-KNOB_INCREMENT_AMOUNT);
```
mcslee: Ditto on `0-` being weird, just use `-`

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
```diff
@@ -337,12 +350,57 @@ public void onParameterChanged(LXParameter parameter) {
...
+        }
+      }
+    }
+
+    private void onKnobIncrement(int index) {
```
mcslee: Feels really repetetive with the other mode. I'm not strictly opposed but a single method with a boolean param for inc vs. dec seems like it'd save code here.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
```diff
@@ -490,9 +548,23 @@ public void controlChangeReceived(MidiControlChange cc) {
...
-            this.deviceListener.onKnob(number - DEVICE_KNOB, cc.getNormalized());
-            return;
+          int iKnob = number - DEVICE_KNOB;
+          if (this.deviceListener.isKnobRelative(iKnob)) {
+            if (note == KNOB_INCREMENT || note == KNOB_INCREMENT_FAST || note == KNOB_INCREMENT_VERYFAST) {
+              this.deviceListener.onKnobIncrement(iKnob);
+            } else if (note == KNOB_DECREMENT || note == KNOB_DECREMENT_FAST || note == KNOB_DECREMENT_VERYFAST) {
+              this.deviceListener.onKnobDecrement(iKnob);
+            } else {
+              // Knob sent absolute values but software is expecting relative values
```
mcslee: Ooh man that's weird... particularly since the note values for INCREMENT, INCREMENT_FAST etc. are all values which could very well be normal knob rotation values. If the MFT is in normal mode, it's gonna go funny and suddenly do increment/decrement things when moving through that 61-67 range.

Is there a way to protect against this? How does setting the MFT mode work?

Also thinking we might want to rename `note` to `value` here... bit confusing as it's not a note number.

src/main/java/heronarts/lx/parameter/DiscreteParameter.java
```diff
@@ -64,6 +64,7 @@ public DiscreteParameter(String label, int value, int min, int max) {
     super(label, value);
     setRange(min, max);
     setUnits(LXParameter.Units.INTEGER);
+    setWrap(true);
```
mcslee: I think this definitely should not be true by default, as the base case for DiscreteParameters is just an integer value range. Don't think those should all wrap if they slam around the end as behavior at 0 vs. MAX may be *very* discontinuously different.

For EnumParameter I would maybe entertain making it wrappable by default, but think mostly as the precedent is already set here, all should be false by default and can be explicitly set true only where desired.

src/main/java/heronarts/lx/parameter/LXListenableNormalizedParameter.java
```diff
@@ -57,4 +58,30 @@ public boolean isMappable() {
     return this.mappable;
   }
 
+  public boolean isWrap() {
+    return this.allowWrap;
+  }
+
+  public LXListenableNormalizedParameter setWrap(boolean allowWrap) {
```
mcslee: I think this stuff should live in the `LXNormalizedParameter` interface. For naming, let's have it mirror `setMappable()` and `isMappable()`, so

So in LXNormalizedParameter:
```
public default boolean isWrappable() {
  return false;
}
```

Then you can leave the setter here in LXListenableNormalizedParameter as I think that will cover all cases where we care about being able to set it to anything else.

mcslee (changes requested):
Overall supportive of this change - makes sense and seems reasonable but I don't love the fact that the MFT mode is opaque. Is there no way to set it via the control surface initialization? What does mode-switching look like?

Had a few other naming suggestions on the isWrappable() stuff. We could possibly go full enum WrapMode there... but I think isWrappable() is fine for now and I can't really imagine any other wrapping modes. Should be default false all the time though I'd say.

As for when to bump a discrete parameter, I think that should probably be an IncrementMode option on DiscreteParameter. I can imagine scenarios where both behaviors are reasonable. Some DiscreteParameters are really just like BoundedParameters that only have int values, but you think of them like a knob and it should just move over the range. Others are more like EnumParameters where you are iterating through settings and it makes send to iterate in that way.


Could be something like this on DiscreteParameter:
```
enum IncrementMode {
  NORMALIZED,
  RELATIVE
}
```

Then the MFT could look at which mode is in use and either do tick-counting (RELATIVE mode) or just move as if it's a bounded knob (e.g. full range of knob covers full normalized range).

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
jkbelcher: Hmm no it didn't.  I created this branch after the commit from my other pending pull request, perhaps that was the wrong move.  What's the best practice when staging multiple pull requests... branch each of them off the current dev branch?  I wasn't sure if doing that would require each of them to be rebased before merge and therefore more clunky on your end.  Still a little green on the git techniques.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: No the knob mode can't be set via MIDI or by button presses.  You have to do it through the Midi Fighter Utility application.  It is a downside, anyone running a MFT needs to configure it for their application... in this case apply the LX-specific settings.  I have notes about the proper settings at the top of the MFT file.

Well... that's not entirely true.  The Midi Fighter Utility writes all the system settings and encoder settings to the device with a long (or maybe multiple?) sysex message.  (system settings + 64 virtual knobs*[11 settings per knob] is a fairly long list.)  So we *could* mimic the device-flash sysex and set all the settings to be LX-compatible on startup.  That seems a bit bold... but OTOH it sure would eliminate any issues with people failing to get the settings right and having issues running them with LX.

Doing that would overwrite a user's previous settings on the MFT... so if they had a custom configuration for some other software they would need to export their settings through the Utility software and re-flash them when switching back.  I guess they would have to do that anyway.  We could try to play nice by reading the device settings on initialization, sending LX settings, and then re-send original settings before shutdown.  No avoiding losing the settings though on a non-clean shutdown.  But hey, it would be slick for people in a stable environment.

mcslee: > We could try to play nice by reading the device settings on initialization, sending LX settings, and then re-send original settings before shutdown. No avoiding losing the settings though on a non-clean shutdown. But hey, it would be slick for people in a stable environment.

If this is possible that would be great. It seems analogous to me to how Ableton puts the APC40 into Live Mode when it's activated as a control surface, and then puts it back afterwards when it's done. If we were able to do the same thing with the MFT, that'd be the most slick approach and would ensure that it's always behaving properly in control surface mode.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
mcslee: Yeah using separate branches off dev is generally the move and then pull-request from each, which contains one set of changes. GitHub will generally be clever about dealing with the rebasing stuff fairly automatically so long as there are no conflicts (which there wouldn't be in this case as they're touching separate files anyways).

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: Got it, done.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: Done.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: Good call.  Will change.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
jkbelcher: Ok, great to know.  Will do it this way in the future.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: Looks like this is achievable.  If it's okay with you can we leave this for a later pull request?  Might not get done before Treefort and could be a mini-project of its own.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: Ok to leave the odd fall-through handling until we get the sysex stuff sorted out?

The string of three ORs does look strange... I kept adding them on as I discovered flicking the knob sends a "higher velocity" relative change.  I could change this to an inclusive range if that would look better?

Renamed note to value, thanks for catching.

src/main/java/heronarts/lx/parameter/DiscreteParameter.java
jkbelcher: Definitely happy to default everything to wrappable=false.  I only made this one true because I was replacing a previously existing method DiscreteParameter.increment(int) which used a hardcoded wrap=true, and didn't want to muck up anyone's behavior out there if it was in use:

https://github.com/heronarts/LX/pull/55/files#diff-4c24b49cb2e2e475b792bb7a1641236cc4b734d713e6ac97231d1add7e96a6bfL188

jkbelcher:
> As for when to bump a discrete parameter, I think that should probably be an IncrementMode option on DiscreteParameter. I can imagine scenarios where both behaviors are reasonable. Some DiscreteParameters are really just like BoundedParameters that only have int values, but you think of them like a knob and it should just move over the range. Others are more like EnumParameters where you are iterating through settings and it makes send to iterate in that way.
> 
> Could be something like this on DiscreteParameter:
> 
> ```
> enum IncrementMode {
>   NORMALIZED,
>   RELATIVE
> }
> ```

Great, was looking for a way to accommodate both types.  Which do you think should be the default?  Maybe default to NORMALIZED on DiscreteParameter and RELATIVE on ObjectParameter<>?

src/main/java/heronarts/lx/parameter/DiscreteParameter.java
mcslee: Oof, okay hadn't looked at that closely and I thought I'd defaulted that to false. If it's true, then I guess you're right we should keep the default true here. I recall now, I did default that to true so that a keyboard up/down arrow on DiscreteParameters will wrap around the end. On a knob this maybe feels a little different than a discrete keyboard press, but I think it's better not to overcomplicate and to keep it consistent with how it was behaving before. So yes let's keep this `setWrappable(true)` for `DiscreteParameter`.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
mcslee: > The string of three ORs does look strange... I kept adding them on as I discovered flicking the knob sends a "higher velocity" relative change. I could change this to an inclusive range if that would look better?

I think that's fine, better to spell out the three explicit values, they are adjacent but not really a range so much as three magical values with similar meanings.

> Ok to leave the odd fall-through handling until we get the sysex stuff sorted out?

Yeah that's fine for now.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
mcslee: Yeah I'm fine to leave that for a later thing.

mcslee:
> Great, was looking for a way to accommodate both types. Which do you think should be the default? Maybe default to NORMALIZED on DiscreteParameter and RELATIVE on ObjectParameter<>?

Yep agreed, that sounds good.

src/main/java/heronarts/lx/parameter/DiscreteParameter.java
jkbelcher: I'm just going to throw one more "are you sure?" in here.  If a normal expected behavior for DiscreteParameter would be not wrappable, this choice would be causing problems for future people in order to avoid problems for past people.

It could be a good compromise to default DiscreteParameter to wrappable(false) and ObjectParameter to wrappable (true).  The default behavior I would expect on keyboard up/down (and everywhere else) would be for EnumParameter to wrap but not a list of 1-100 integers.

mcslee: Okay cool - yeah I'm agreed, let's kill the keyboard wrapping too, does seem perfectly reasonable not to expect that. Wrappable by default just for ObjectParameter/EnumParameter it is!

src/main/java/heronarts/lx/parameter/LXListenableNormalizedParameter.java
jkbelcher: Adjusted.  See if it looks right after I push changes.

jkbelcher:
Ok, updates available for review.  Had to add some special handling of DiscreteParameter when IncrementMode=NORMALIZED on the MFT because you can't just bump the normalized value along to the next value.  I went with a solution similar to UIKnob where the controlling device tracks an absolute value for the parameter.

src/main/java/heronarts/lx/parameter/LXListenableNormalizedParameter.java
```diff
@@ -57,4 +58,31 @@ public boolean isMappable() {
...
+    return this;
+  }
+
+  @Override
+  public boolean isWrappable() {
+    return this.wrappable;
+  }
+
+  public LXListenableNormalizedParameter incrementNormalized(double amount) {
+    return incrementNormalized(amount, this.wrappable);
```
mcslee, line 72: Should use `isWrappable()` here in case a subclass were to override it... probably a non-issue, I'll quickly fix post-merge.

mcslee:
Made some very minor post-commit touchups: https://github.com/heronarts/LX/commit/c259d40f2be5d275c8fba0b116ef66672207c0a1

PR #56, opened 2022-03-15
Add AlternatingPattern

raldi:


LICENSE
```diff
@@ -1,4 +1,4 @@
-Copyright 2013-2022 Mark C. Slee, Heron Arts LLC		
+Copyright 2013-2022 Mark C. Slee, Heron Arts LLC
```
mcslee: Not a big deal but mind knocking this off the PR?

src/main/java/heronarts/lx/LXRegistry.java
```diff
@@ -113,6 +113,7 @@ public LXRegistry removeListener(Listener listener) {
     DEFAULT_PATTERNS = new ArrayList<Class<? extends LXPattern>>();
     DEFAULT_PATTERNS.add(heronarts.lx.pattern.color.GradientPattern.class);
     DEFAULT_PATTERNS.add(heronarts.lx.pattern.color.SolidPattern.class);
+    DEFAULT_PATTERNS.add(heronarts.lx.pattern.color.AlternatingPattern.class);
```
mcslee: Let's leave this out of here for now. You can register the pattern for use in your project in your `initialize()` callback method with `lx.registry.addPattern()` - I'm trying to keep the core set of patterns/effects pretty basic and polished. Don't think the Alternating pattern does enough to warrant it just yet, but maybe after some more options and modes are added.

src/main/java/heronarts/lx/color/LinkedColorParameter.java
```diff
@@ -76,6 +76,23 @@ protected void onSubparameterUpdate(LXParameter p) {
     }
   }
 
+  // Returns the real-time value of the color, which may be different from what
+  // getColor() returns if there are LFOs/etc being applied.
+  public int calcColor() {
```
mcslee, line 81: Nice, great addition here

src/main/java/heronarts/lx/pattern/color/AlternatingPattern.java
```diff
@@ -0,0 +1,75 @@
...
+    this.firstColor.setColor(firstColor);
+    this.secondColor.setColor(secondColor);
+    addParameter("firstColor", this.firstColor);
+    addParameter("secondColor", this.secondColor);
+    addParameter("stripeLength", this.stripeLength);
+  }
+
+  @Override
+  public void run(double deltaMs) {
+    LinkedColorParameter[] colorParameters = {this.firstColor, this.secondColor};
```
mcslee: This and the line below are going to allocate new dynamic arrays on every single iteration of the run loop that will need garbage collecting. Best to avoid that wherever possible. If you really want to use arrays here you should make them member variables on the class. But honestly, with just two colors here, I think the below is more readable. Also recommend pre-fetching a parameter value if it's going to be used many times... otherwise it's N function calls for a N-pixel project... the JVM can sort of optimize-it-away and CPU is plentiful these days, but best to avoid 50K+ function calls where you can.

```
int stripeLength = this.stripeLength.getValuei();
int color1 = this.firstColor.calcColor();
int color2 = this.secondColor.calcColor();
for (int i = 0; i < colors.length; ++i) {
    int phase = (i / this.stripeLength) % 2;
    this.colors[i] = (phase == 0) ? color1 : color2;
}
```

src/main/java/heronarts/lx/pattern/color/SolidPattern.java
```diff
@@ -33,6 +33,8 @@
 
   public SolidPattern(LX lx) {
     this(lx, LXColor.RED);
+    this.color.mode.setValue(LinkedColorParameter.Mode.PALETTE);
```
mcslee: Let's hold off on this default-change for now, I think for most non-power users with stock LX sticking with a basic direct color as default is best.

If this is driving you bonkers having to change all the time, you can make your own subclass of SolidColorPattern for your project which flips these defaults at construction-time and register that in your initialize(), same as Alternating

src/main/java/heronarts/lx/pattern/color/SolidPattern.java
```diff
@@ -43,19 +45,6 @@ public SolidPattern(LX lx, int color) {
...
-    case STATIC:
-      // There may be modulators applied to the h/s/b values!
-      setColors(LXColor.hsb(
-        this.color.hue.getValue(),
-        this.color.saturation.getValue(),
-        this.color.brightness.getValue()
-      ));
-      break;
-    }
+    setColors(color.calcColor());
```
mcslee, line 46: Nice, much better.

mcslee (changes requested):
Nice, looks good - just a few minor things and I would rather leave the LX defaults in place for now.

LICENSE
raldi: Trailing whitespace restored.  ;)

src/main/java/heronarts/lx/LXRegistry.java
raldi: Done

src/main/java/heronarts/lx/pattern/color/AlternatingPattern.java
raldi: Ah, I overestimated the sophistication of the optimizer. Good to know. Updated to use your code.

src/main/java/heronarts/lx/pattern/color/SolidPattern.java
raldi: Done

raldi:
PTAL

src/main/java/heronarts/lx/pattern/color/AlternatingPattern.java
mcslee: It is something that I'm always curious about. It *might* be able to figure out those arrays are not-really-dynamic and reuse them, or at least they will just get smartly re-allocated when the next loop comes around asking for an array of the same size. I shouldn't speak definitively about this, as I don't really know how it works. But I try not to count on that sort of magic in cases where it is easily avoided, particularly inside the run() loops.

PR #57, opened 2022-03-17
Improve cue control on APC40mkII

jkbelcher:
As discussed, except added a flag to track a fork in behavior.  For the first button press down, if the cue was active, it could be an un-cue or the start of a multi-cue.  For the start of the multi-cue, it feels right to be able to start with an already cued channel and add other ones to it.  So for a single button press on an active cue, the un-cue doesn't happen until button up.  Tested, seems logical on the board.

Happy to rename that variable - left it verbose so the behavior was obvious.


mcslee:
Nice - FYI did a few cleanups here, no functionality change but you had a couple if/else blocks that could be collapsed into single lines.

https://github.com/heronarts/LX/commit/6dd5df6ddf3aa3d472b30f9c1aeebfd8f6eba14e

jkbelcher:
Thanks, looking better now.

PR #58, opened 2022-03-22
APCMk2 uses one physical knob for each LinkedColor

raldi:


src/main/java/heronarts/lx/color/ColorParameter.java
```diff
@@ -43,6 +43,7 @@ public ColorParameter(String label, int color) {
     double h = LXColor.h(color);
     this.hue = new CompoundParameter("Hue", Double.isNaN(h) ? 0 : h, 0, 359)
       .setDescription("Hue component of the color");
+    this.hue.setWrappable(true);
```
mcslee, line 46: Nice.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -216,25 +217,49 @@ private void updateGridMode() {
     sendChannelFocus();
   }
 
+  // When turning a knob that sometimes controls Hue, sometimes Brightness,
+  // sometimes Saturation, look up the current mode of the surface and return
+  // the appropriate subparameter that this knob currently controls.
+  private CompoundParameter getColorSubparam(ColorParameter colorParameter) {
+    if (this.shiftOn) {
+      if (this.bankLeftOn || this.bankRightOn) {
```
mcslee, line 234: Feels sort of weird that in shift-mode, both bank buttons turn it to hue and nothing does brightness, but I guess I see no problem with that.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -265,6 +290,7 @@ void registerChannel(LXBus channel) {
       } else {
         register(null);
       }
+      updateLCPKnobs();
```
mcslee: It feels like this should be handled within register() as it loops over knob construction.. Is it having another side effect that I'm not noticing here?

src/main/java/heronarts/lx/parameter/LXListenableParameter.java
```diff
@@ -39,7 +39,7 @@
 
   private final List<LXParameterListener> listeners = new ArrayList<LXParameterListener>();
 
-  private AggregateParameter parentParameter = null;
+  public AggregateParameter parentParameter = null;
```
mcslee: Use getParentParameter() which was defined on LXParameter.java... setting it to public opens a hole in the API where someone could improperly modify its value.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -216,25 +217,49 @@ private void updateGridMode() {
...
-    private final LXListenableNormalizedParameter[] knobs =
-      new LXListenableNormalizedParameter[DEVICE_KNOB_NUM];
+    private final LXListenableParameter[] knobs =
+            new LXListenableParameter[DEVICE_KNOB_NUM];
 
     DeviceListener() {
-      for (int i = 0; i < this.knobs.length; ++i) {
-        this.knobs[i] = null;
-      }
+      Arrays.fill(this.knobs, null);
```
mcslee, line 264: Oh my god how did I never know about that helper. Thanks for this!

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -291,50 +317,88 @@ void registerNext() {
...
-              sendControlChange(0, DEVICE_KNOB + i, (int) (normalized * 127));
-            } else {
-              sendControlChange(0, DEVICE_KNOB_STYLE + i, LED_STYLE_OFF);
-            }
-            ++i;
+        if (i >= this.knobs.length) {
+          // Subtle: Even if all the knobs are full, we need to keep adding
+          // listeners for all children of the final knob. So only break
+          // if we're up to a parameter with no parent or a different parent.
+          if (parameter.parentParameter == null || parameter.parentParameter != currentParentParameter) {
```
mcslee: `AggregateParameter parentParameter = parameter.getParentParameter();`

Grab that once via accessor and use it for all the references from here down.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -291,50 +317,88 @@ void registerNext() {
...
+      boolean isEnabled = false;
+      if (this.effect != null) {
+        isEnabled = this.effect.isEnabled();
+      } else if (this.pattern != null) {
+        isEnabled = this.pattern == ((LXChannel) this.channel).getActivePattern();
+      }
+      sendNoteOn(0, DEVICE_ON_OFF, isEnabled ? LED_ON : LED_OFF);
+
+      int i = 0;
+      LXListenableParameter currentParentParameter = null;
```
mcslee: I'm not sure this approach is totally safe, as `getRemoteControls()` is not required to return all the children of an `AggregateParameter` immediately after it. It will do that by default, but the idea with getRemoteControls() is that the implementor of a Pattern/Effect can specify which parameters they want to be used by a control surface, and in which order. Doesn't necessarily have to map to how they were declared.

I think your loop below that goes past the end of the knobs.length is relying up on the assumption that once the "currentParentParameter" value changes, we're done with those sub-params. But that is not necessarily the case.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -291,50 +317,88 @@ void registerNext() {
...
-              double normalized = (parameter instanceof CompoundParameter) ?
-                ((CompoundParameter) parameter).getBaseNormalized() :
-                parameter.getNormalized();
-              sendControlChange(0, DEVICE_KNOB + i, (int) (normalized * 127));
-            } else {
-              sendControlChange(0, DEVICE_KNOB_STYLE + i, LED_STYLE_OFF);
-            }
-            ++i;
+        if (i >= this.knobs.length) {
+          // Subtle: Even if all the knobs are full, we need to keep adding
```
mcslee: Just checking here... why do we need listeners for the children of the final knob? `AggregateParameter` should be a Listenable parameter, so listening to that ought to pick up the relevant changes that would make this knob have to change?

Wondering if perhaps the issue here is that I am not always firing a change event on LinkedColorParameter subparam changes, if the color isn't actually modified? Is that what this is playing defense against?

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -291,50 +317,88 @@ void registerNext() {
...
+            currentParentParameter = parameter.parentParameter;
+            knobParam = currentParentParameter;
+          } else {
+            // This parameter is the non-first child to a parent.
+            // The parent's was already attached to a knob when
+            // we processed its oldest sibling.
+            continue;
+          }
         }
+        this.knobs[i] = knobParam;
```
mcslee: Won't this array overflow if `i >= this.knobs.length` but the checks above don't `break` or `continue`?

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -349,17 +413,39 @@ public void onParameterChanged(LXParameter parameter) {
...
+    // is trivial except for LinkedColorParameters, whose function changes depending
+    // on the surface's mode.
+    private LXListenableNormalizedParameter parameterForKnob(LXListenableParameter knob) {
+      if (knob == null || knob instanceof LXListenableNormalizedParameter) {
+        // If the knob is unused or linked to a vanilla parameter, we're done.
+        return (LXListenableNormalizedParameter) knob;
+      }
+
+      // The knob is attached to an LinkableColorParameter.
+      LinkedColorParameter lcp = (LinkedColorParameter) knob;
```
mcslee: Need an explicit check here for `knob instanceof LinkedColorParameter`

If someone passes in a basic `ColorParameter` or some other kind of `AggregateParameter` this is going to throw a `ClassCastException`

I think we should also make this work for `ColorParameter` - not just `LinkedColorParameter` - in that case it would be nice for it to do the hue/sat/brightness magic with the bank buttons, without the ability to select a swatch color.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -370,9 +456,46 @@ void onDeviceOnOff() {
...
+
+        if (focusColor != null) {
+          int palIndex = lx.engine.palette.swatch.colors.indexOf(focusColor);
+          if (palIndex >= 0) {
+            lcp.mode.setValue(LinkedColorParameter.Mode.PALETTE);
+            lcp.index.setValue(palIndex + 1);
+            return;
+          }
+        }
+        if (colorClipboard != null) {
```
mcslee, line 474: Similarly here, I think this part should work for a basic `ColorParameter` as well, where the color value can be copied without linking.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -370,9 +456,46 @@ void onDeviceOnOff() {
...
+      }
+
+      LXListenableNormalizedParameter knobParam = parameterForKnob(knob);
+
+      // If it's a wrappable parameter, let it wrap
+      if (knobParam.isWrappable()) {
+        if (normalized == 0.0) {
+          normalized = 1.0;
+        } else {
+          normalized = (normalized + 1.0) % 1.0;
```
mcslee: This is sort of confusing to read, I think might be clearer if you had an `else if (normalized == 1.0)` set it to 0. With a comment, that would make clear that this is catching the cases where the knob has just spun UP to 1., or DOWN to 0.

This mod-math expression *only* actually results in a different value for `1.0`, which gets changed to `0.0` - but this is not immediately clear from looking at it. I think parity with the above expression turning 0.0 to 1.0 will make this more understandable.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -381,12 +504,19 @@ private void unregister() {
       }
       if (this.device != null) {
         for (int i = 0; i < this.knobs.length; ++i) {
-          if (this.knobs[i] != null) {
+          if (this.knobs[i] == null) {
+            continue;
+          } else if (this.knobs[i] instanceof AggregateParameter) {
+            AggregateParameter ap = (AggregateParameter) this.knobs[i];
+            for (LXListenableParameter sub : ap.subparameters.values()) {
```
mcslee, line 520: Per the above comments about getRemoteControls() - there is no guarantee here that we have registered as a listener to every sub-parameter of an `AggregateParameter`, they may or may not have all been returned by getRemoteControls().

I think the safest solution here is that we should only listen to parameters directly registered as knobs, and we need to make sure that any `AggregateParameter` does properly fire a listener message when something which changes its display has happened.

Otherwise I worry that this extra-listener management could get brittle.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -834,6 +975,30 @@ private void sendRainbowPickerGrid() {
...
 
+  // For each knob controlling a LinkedColorParameter, update its lighting to match
+  // the value it would change if turned right now. If we ever have an
+  // AggregateParameter knob where some subparameters are unipolar and others
+  // bipolar, we'll need to do a sendControlChange() as well here. For now,
+  // though, that never happens.
+  private void updateLCPKnobs() {
+    for (int i = 0; i < this.deviceListener.knobs.length; i++) {
+      LXListenableParameter knob = this.deviceListener.knobs[i];
+      if (knob instanceof LinkedColorParameter) {
```
mcslee: Can also handle basic `ColorParameter` here. I think method name of `updateColorKnobs()` would help for clarity.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -834,6 +975,30 @@ private void sendRainbowPickerGrid() {
...
+        LinkedColorParameter lcp = (LinkedColorParameter) knob;
+
+        LXListenableNormalizedParameter subparam;
+
+        if (lcp.mode.getEnum() == LinkedColorParameter.Mode.PALETTE) {
```
mcslee: This feels redundant with `parameterForKnob`

mcslee (changes requested):
Overall looks like a good structure, but there are a couple core bits I think need ironing out.

Specifically, getRemoteControls() offers no guarantee about which parameters are returned or in what order. There are a bunch of assumptions in some of your loops about sub-params coming immediately after an AggregateParameter. That's not guaranteed, and this needs to work whether or not that is the case. I think it would be best if we don't rely upon listening to sub-params, but just listen to the AggregateParameter itself for changes and update accordingly.

The other big thing is that I think all of this stuff should work for a basic `ColorParameter`, with the functionality enhanced to support linking when it's a `LinkedColorParameter`

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: It makes more sense if you think of it like, in either bank mode, holding shift momentarily disables the bank mode and makes the knob do its normal thing (hue), and for backwards compatibility with existing UX, holding shift when you're not in a bank mode momentarily makes the knob control saturation.

src/main/java/heronarts/lx/parameter/LXListenableParameter.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: IntelliJ suggested it; all I did was make a single click.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Oh, I didn't realize that. Done ... but it's too bad that by the time onKnob() is called, the value is already normalized (truncated) because it means if the current value is 0.99 and they give it a big turn that would normally be worth an adjustment of +0.25, the new value is 0.0 instead of 0.24.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: It needs to be callable elsewhere, too: When shift, bankLeft, or bankRight are pressed, we need to re-light the knobs according to the underlying value they newly control.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: I've refactored since you made this, but at least in the current version, the only way we get past the `if (i >= this.knobs.length)` check is if `parent` is non-null and the same as `currentParent`, but in that case, the `continue` is hit.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Let's say we've filled up 7 knobs and the next parameter to be processed is the Hue of a new LCP. We assign the LCP to the 8th knob and set up a listener for the Hue parameter, so that when it changes later, we can update the knob's lights. But we need to do the same thing with saturation and brightness. Without this special case, on the next two loops when we're looking at the S and B params, we'd skip adding listeners because the 8 knobs are full.

raldi (commented):
Addressed everything except the comments on parameters coming in out of order. I can't reproduce this, or artificially trigger it, and I'm wary of implementing a solution until I can observe the failure case. Can you describe a way I can bring about the out-of-order scenario you descibe?

raldi:
Hmm, this isn't working quite right. Since changing to cc.getNormalized(), changing saturation through the CUE_LEVEL knob doesn't seem to work anymore. Will investigate...

mcslee:
FYI getting on a plane so I’ll be a bit slow to reply to this next revision.

For the controls issue, make a test pattern with a few parameters and then
call in it’s constructor:

setRemoteControls(
  p1,
  p2,
  p3,
  p4,
  p5,
  lcp1.hue,
  lcp1.saturation,
  lcp2.hue,
  lcp1.brightness,
  p6,
  lcp2.mode
);

p1-6 can be whatever. But the point is to overflow the number of knobs and
for the sub-parameters of the linked color parameters to not all be present
and to be out of order, particularly in slots 8-9 where we’ll hit
knobs.length.

raldi:
Ah, I hadn't realized that was something a pattern could customize. I'll
take a crack at this tomorrow.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: I think I've addressed all of your feedback except for this one. I think I need a bit more detail if I'm going to make this change to AggregateParameter.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
raldi: Done

raldi:
I think everything is ready now.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -216,25 +214,61 @@ private void updateGridMode() {
...
+      LinkedColorParameter lcp = (LinkedColorParameter) agg;
+      if (lcp.mode.getEnum() == LinkedColorParameter.Mode.PALETTE) {
+        // If it's palette-linked, the knob always controls the palette index.
+        return lcp.index;
+      }
+    }
+
+    // So far, ColorParameters are the only kind of AggregateParameter. If
+    // we add more later, we'll have to add custom code here to handle them.
+    ColorParameter colorParameter = (ColorParameter) agg;
```
mcslee, line 231: This should still check instanceof ColorParameter and return a null for subparameter if it doesn't recognize the AggregateParameter type... otherwise it'll throw an Exception and crash the whole thing.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -291,50 +326,79 @@ void registerNext() {
...
+
+      this.device.controlSurfaceSemaphore.increment();
+
+      boolean isEnabled = false;
+      if (this.effect != null) {
+        isEnabled = this.effect.isEnabled();
+      } else if (this.pattern != null) {
+        isEnabled = this.pattern == ((LXChannel) this.channel).getActivePattern();
+      }
+      sendNoteOn(0, DEVICE_ON_OFF, isEnabled ? LED_ON : LED_OFF);
```
mcslee, line 361: Think this needed to also happen in the `(this.device == null)` case.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -291,50 +326,79 @@ void registerNext() {
...
+
+      int i = 0;
+      Set<AggregateParameter> knownParents = new HashSet<>();
+      for (LXListenableNormalizedParameter parameter : this.device.getRemoteControls()) {
+        if (parameter == null) {
```
mcslee, line 366: Feels like this can still throw a `NullPointerException` and crash if there is an explicit `null` in the list of knobs while `i >= this.knobs.length`

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -291,50 +326,79 @@ void registerNext() {
...
-              sendControlChange(0, DEVICE_KNOB + i, (int) (normalized * 127));
-            } else {
-              sendControlChange(0, DEVICE_KNOB_STYLE + i, LED_STYLE_OFF);
-            }
-            ++i;
-          }
-          this.device.controlSurfaceSemaphore.increment();
+        AggregateParameter parent = parameter.getParentParameter();
+
+        if (knownParents.contains(parent) || i >= this.knobs.length) {
```
mcslee, line 374: Seems like we should just be `break`-ing on `i >= this.knobs.length` - how it was before. We'll never get past this `continue` once i reaches that cutoff, and all we've got above it is the chance to crash on a `null` parameter...

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -349,17 +413,33 @@ public void onParameterChanged(LXParameter parameter) {
...
+    // is trivial except for LinkedColorParameters, whose function changes depending
+    // on the surface's mode.
+    private LXListenableNormalizedParameter parameterForKnob(LXListenableParameter knob) {
+      if (knob == null || knob instanceof LXListenableNormalizedParameter) {
+        // If the knob is unused or linked to a vanilla parameter, we're done.
+        return (LXListenableNormalizedParameter) knob;
+      }
+
+      // The knob is attached to an AggregateParameter.
+      AggregateParameter agg = (AggregateParameter) knob;
```
mcslee, line 438: Even though I can't think of a case that would be something else here... better to code this defensively with an `instanceof` check to avoid `ClassCastException` crashing the whole program.

src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
```diff
@@ -381,12 +513,19 @@ private void unregister() {
       }
       if (this.device != null) {
         for (int i = 0; i < this.knobs.length; ++i) {
-          if (this.knobs[i] != null) {
+          if (this.knobs[i] == null) {
+            continue;
+          } else if (this.knobs[i] instanceof AggregateParameter) {
+            AggregateParameter ap = (AggregateParameter) this.knobs[i];
```
mcslee, line 519: Need to duplicate the Set technique here for seenKnobs... this won't crash the program but it will cause a lot of error log spew about trying to unregister listeners that don't exist. Because if multiple knobs are on the same aggregate parameter, the first will unregister all the listeners, then there will be subsequent redundant calls to removeListener() after it's already gone.

mcslee (approved):
Looks generally good. Still some minor issues but for the sake of unblocking this going to merge and I'll follow up quickly here with some cleanups.

mcslee:
Here's a bit of cleanup:
https://github.com/heronarts/LX/commit/8e92758affdb2ba3839cddff11a9b4aab712fe78

Some of the checks there are a bit more robust. The main functionality difference was that your implementation would skip adding knobs the second time the same parent parameter was seen. But I don't think we should do this - the idea of getRemoteControls() is that the programmer can specify a specific layout for the ordering of knobs on a control surface (in the future potentially across multiple dynamic pages). This is why `null` is expressly allowed in that list, if someone wants to say "I really want this parameter in spot 7, even though there's nothing in knob spots 5 and 6" - e.g. for ergonomic  layout reasons.

I think that's the right general behavior, but if this is super awkward for your project resulting in loads of redundant color knobs... I would recommend you make a little helper for your pattern classes that calls setRemoteControls() and filters out adding all the ColorParameter sub-params by default.

Also removed that one updateColorKnobs() call in registerChannel - as I think that was better handled by having register() properly set the value on each color knob at init-time, which it now does.

I was flying slightly blind on this as don't have an APC40mk2 to test here - so you'll have to final test for now until I'm back home where there's one on my desk. Apologies if something was not *quite* right in here...

raldi:
I agree with you about honoring setRemoteControls() for devices that call it, but right now the default for patterns that don't is that each of their LinkedColorParameters, if any, take up five physical knobs on the APC. I can write a helper function for our patterns, but wouldn't it make more sense to set things up so that the default was that each LCP gets one physical knob, and if someone wants to customize, they can use setRemoteControls()? I think I can write a PR that sets things up that way, if you agree.

mcslee:
Yeah, agreed. I thought about doing this while I was doing the cleanup change-set, but then shrugged it off out of laziness/is-it-really-necessary/it-was-late. Now you've said the same thing, so I'm sold.

https://github.com/heronarts/LX/commit/d19eb4902eb1696aefa3de2879b8342693e2d5ac

I made this something that AggregateParameter classes can specify. Let me know if this works as expected for you?

PR #59, opened 2022-03-25
Refactoring performed on the code

janvipatel1412:
I went over the project and refactored a few items to make the code more readable.
The changes to the code are:

1. Rename Variable:
In the class StrobeEffect.java, I refactored a variable that was declared as amt to amount to make it more meaningful.

2. Extract Method:
In the class FourierTransform.java, I refactored a method named computing. I extracted a block of code from the method as it was feasible to put a logic that was different from another method.

3. Move Variable:
In the class SparkleEffect.java, a variable amount was more used in SparklePattern.java rather than the SparkleEffect.java class. So, it was moved to another class.

4. Pull up variable:
A variable was not declared in the class LXAudioComponent.java, but its two subclasses LXAudioInput.java and LXAudioOuput.java had it declared. So, I removed the variables from both the subclasses and pulled them up to the superclass.

5. Extract Class:
In the class LXCommand.java, there were other subclasses that performed different operations. So, I moved them to another class which I made.


src/main/java/heronarts/lx/audio/FourierTransform.java
```diff
@@ -137,14 +137,38 @@ public FourierTransform compute(float[] samples) {
       this.real[i] = samples[bri] * this.windowCoefficients[bri];
       this.imaginary[i] = 0f;
     }
+
+    computeComplexNum();
```
mcslee, line 141: This doesn't really seem like an improvement to me... the function isn't called anywhere else, and for someone to follow the whole algorithm they now read code above this, jump down to read the function below, then come back up to read the code after the function call. If it was called elsewhere I think that'd be sensible to decompose, but as it's not, I think this is better inline.

src/main/java/heronarts/lx/audio/LXAudioComponent.java
```diff
@@ -52,6 +52,8 @@
   public final LXAudioBuffer right = new LXAudioBuffer(SAMPLE_BUFFER_SIZE, SAMPLE_RATE);
   public final LXAudioBuffer mix = new LXAudioBuffer(SAMPLE_BUFFER_SIZE, SAMPLE_RATE);
 
+  protected boolean stopped = false;
```
mcslee, line 55: I don't think this helps. The variable is never referenced in LXAudioComponent, but it has now been downgraded from `private` to `protected` - which opens it up to being touched in ways that could introduce bugs or side-effects in the subclasses.

Each subclass used its `stopped` variable totally internally, which is why they were both `private`

src/main/java/heronarts/lx/command/LXCommand.java
```diff
@@ -1759,177 +1759,7 @@ public boolean isIgnored() {
...
-
-      @Override
-      public String getDescription() {
-        return "Recall Snapshot";
-      }
-    }
-
-    private static class RemoveView extends LXCommand {
-
+     private static class RemoveView extends LXCommand {
```
mcslee, line 1762: These classes were specifically all inside `LXCommand` so that when referenced elsewhere there is a consistent class naming scheme that makes it clear what type of command it is.

e.g. `LXCommand.Snapshots.RemoveView` rather than `RemoveView`

This change arbitrarily moves just a few of many different commands out of this naming hierarchy. I think it becomes particularly less readable for a class like `Update` - which is quite semantically clear as `LXCommand.Snapshot.Update` (this updates a snapshot), vs. someone reading `new Update` is going to really wonder what this means.

These were all grouped as such for that reason.

src/main/java/heronarts/lx/effect/SparkleEffect.java
```diff
@@ -31,13 +31,11 @@
...
 
-  public final CompoundParameter amount =
-    new CompoundParameter("Amount", 1)
-    .setDescription("Amount of sparkle to apply");
+
 
   public SparkleEffect(LX lx) {
     super(lx);
-    addParameter("amount", this.amount);
+    addParameter("amount", engine.amount1);
```
mcslee, line 38: Renaming this to `amount1` feels pretty clunky, and the parameter was intentionally on `SparkleEffect` because it is only registered as a parameter to `SparkleEffect` using the `addParameter` method, which makes it show up as a UI control for the effect, not the pattern. Hence, the parameter is declared and lives in the effect class.

I can see how the amount-passing is awkward here though, so have changed that to act as a passed parameter: https://github.com/heronarts/LX/commit/a05cfd14a14cf52da254c417fd33591d163dd873

mcslee (commented):
Nice to see you diving into the LX code, but to honest this feels like touching and moving around small bits of a bunch of unrelated things for the sake of touching stuff.

I made one cleanup change to clarify the Sparkle stuff (https://github.com/heronarts/LX/commit/a05cfd14a14cf52da254c417fd33591d163dd873), but the rest of this I don't think warrants changing.

Just curious - are you working on an active LX project? I'm just curious what motivated this, and happy to take a look at what you're using LX for. I'm going to close out this request, but happy to hear more about what you're up to. Cheers!

PR #65, opened 2022-07-11
Add input checking on OSC tempo setBPM handler

worldveil:
A couple problems this addresses for real-world deployments:

- Occasionally an OSC source will relay garbage BPM messages (ie: bpm=0.0) and currently the LX engine has no way of ignoring these bogus values
- Applications using LX may have different ideas of what a valid range of BPMs for their usecase are

This PR alleviates these issues by adding sanity checking to new setBPM OSC messages, and allows users to adjust manually the range of allowable BPMs the engine can be set to.

### DIFF
--- a/src/main/java/heronarts/lx/Tempo.java
+++ b/src/main/java/heronarts/lx/Tempo.java
@@ -204,7 +204,7 @@ public class Tempo extends LXModulatorComponent implements LXOscComponent {
   private static final String PATH_SET_BPM = "setBPM";
   
   public boolean setBpmRange(double min, double max) {
-    if (min >= 0.0 && min >= max) {
+    if (min <= 0.0 || min >= max) {
       // do not set to invalid range of BPMs
       return false;
     }
--- a/src/main/java/heronarts/lx/Tempo.java
+++ b/src/main/java/heronarts/lx/Tempo.java
@@ -52,8 +52,8 @@ public class Tempo extends LXModulatorComponent implements LXOscComponent {
   public final static double DEFAULT_MIN_BPM = 20;
   public final static double DEFAULT_MAX_BPM = 240;
 
-  private double minBpm = DEFAULT_MIN_BPM;
-  private double maxBpm = DEFAULT_MAX_BPM;
+  private double minOscBpm = DEFAULT_MIN_BPM;
+  private double maxOscBpm = DEFAULT_MAX_BPM;
 
   public static enum Division {
 
@@ -135,7 +135,7 @@ public class Tempo extends LXModulatorComponent implements LXOscComponent {
     .setDescription("Beats per measure");
 
   public final BoundedParameter bpm =
-    new BoundedParameter("BPM", DEFAULT_BPM, this.minBpm, this.maxBpm)
+    new BoundedParameter("BPM", DEFAULT_BPM, this.minOscBpm, this.maxOscBpm)
     .setDescription("Beats per minute of the master tempo object");
 
   public final BooleanParameter trigger =
@@ -199,18 +199,18 @@ public class Tempo extends LXModulatorComponent implements LXOscComponent {
   private static final String PATH_BEAT = "beat";
   private static final String PATH_SET_BPM = "setBPM";
   
-  public Tempo setBpmRange(double min, double max) {
+  public Tempo setOscBpmRange(double min, double max) {
     if (min <= 0.0 || min >= max) {
       // do not set to invalid range of BPMs
       throw new IllegalArgumentException("Tried to set invalid bpm range!");
     }
-    this.minBpm = min;
-    this.maxBpm = max;
+    this.minOscBpm = min;
+    this.maxOscBpm = max;
     return this;
   }
 
-  public boolean isValidBpm(double bpm) {
-    return bpm >= this.minBpm && bpm <= this.maxBpm;
+  public boolean isValidOscBpm(double bpm) {
+    return bpm >= this.minOscBpm && bpm <= this.maxOscBpm;
   }
 
   @Override
@@ -218,7 +218,7 @@ public class Tempo extends LXModulatorComponent implements LXOscComponent {
     if (parts[index].equals(PATH_SET_BPM)) {
       if (message.size() > 0) {
         float newBpm = message.getFloat();
-        if (isValidBpm(newBpm)) {
+        if (isValidOscBpm(newBpm)) {
           this.bpm.setValue(newBpm);
         }
         return true;
--- a/src/main/java/heronarts/lx/Tempo.java
+++ b/src/main/java/heronarts/lx/Tempo.java
@@ -49,8 +49,11 @@ import heronarts.lx.parameter.MutableParameter;
  */
 public class Tempo extends LXModulatorComponent implements LXOscComponent {
 
-  public final static double MIN_BPM = 20;
-  public final static double MAX_BPM = 240;
+  public final static double DEFAULT_MIN_BPM = 20;
+  public final static double DEFAULT_MAX_BPM = 240;
+
+  private double minBpm;
+  private double maxBpm;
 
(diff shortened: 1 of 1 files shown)


src/main/java/heronarts/lx/Tempo.java
```diff
@@ -191,16 +194,38 @@ public Tempo(LX lx) {
...
+    // set initial minBpm, maxBpm
+    minBpm = DEFAULT_MIN_BPM;
+    maxBpm = DEFAULT_MAX_BPM;
   }
 
   private static final String PATH_BEAT = "beat";
   private static final String PATH_SET_BPM = "setBPM";
+  
+  public boolean setBpmRange(double min, double max) {
+    if (min >= 0.0 && min >= max) {
```
raldi: The specifics of this if-clause surprise me. It's okay for min to be greater than max if they're both negative?

raldi: And it's not okay for min to equal max?

worldveil: good catch. oops, this is opposite of what I intended. fixing

src/main/java/heronarts/lx/Tempo.java
```diff
@@ -49,8 +49,11 @@
  */
 public class Tempo extends LXModulatorComponent implements LXOscComponent {
 
-  public final static double MIN_BPM = 20;
-  public final static double MAX_BPM = 240;
+  public final static double DEFAULT_MIN_BPM = 20;
+  public final static double DEFAULT_MAX_BPM = 240;
+
+  private double minBpm;
```
mcslee: I would just initialize this here:

```
private double minBpm = DEFAULT_MIN_BPM;
private double maxBpm = DEFAULT_MAX_BPM;
```

src/main/java/heronarts/lx/Tempo.java
```diff
@@ -132,7 +135,7 @@ public interface Listener {
     .setDescription("Beats per measure");
 
   public final BoundedParameter bpm =
-    new BoundedParameter("BPM", DEFAULT_BPM, MIN_BPM, MAX_BPM)
+    new BoundedParameter("BPM", DEFAULT_BPM, DEFAULT_MIN_BPM, DEFAULT_MAX_BPM)
```
mcslee: Then just set this to `this.minBpm, this.maxBpm`

src/main/java/heronarts/lx/Tempo.java
```diff
@@ -191,16 +194,38 @@ public Tempo(LX lx) {
     addParameter("trigger", this.trigger);
     addParameter("enabled", this.enabled);
     addModulator("nudge", this.nudge);
+
+    // set initial minBpm, maxBpm
+    minBpm = DEFAULT_MIN_BPM;
```
mcslee: Then you can kill those lines

src/main/java/heronarts/lx/Tempo.java
```diff
@@ -191,16 +194,38 @@ public Tempo(LX lx) {
...
+      return false;
+    }
+    
+    minBpm = min;
+    maxBpm = max;
+    return true;
+  }
+
+  public boolean isValidBpm(double bpm) {
+    return bpm >= minBpm && bpm <= maxBpm;
```
mcslee: LX coding style nitpick - I'm obnoxious about always using `this.minBpm` and `this.maxBpm` to refer to member variables

src/main/java/heronarts/lx/Tempo.java
```diff
@@ -191,16 +194,38 @@ public Tempo(LX lx) {
...
+    minBpm = DEFAULT_MIN_BPM;
+    maxBpm = DEFAULT_MAX_BPM;
   }
 
   private static final String PATH_BEAT = "beat";
   private static final String PATH_SET_BPM = "setBPM";
+  
+  public boolean setBpmRange(double min, double max) {
+    if (min <= 0.0 || min >= max) {
+      // do not set to invalid range of BPMs
```
mcslee, line 204: I would actually `throw new IllegalArgumentException` here - caller probably not checking for boolean return value here, and there shouldn't be any scenario in which passing invalid arguments to this method is sensible.

Can make the method void, or set it's return type to `Tempo` and have it `return this;` for method chaining.

src/main/java/heronarts/lx/Tempo.java
```diff
@@ -191,16 +194,38 @@ public Tempo(LX lx) {
...
+      return false;
+    }
+    
+    minBpm = min;
```
mcslee: Nitpick - `this.minBpm` / `this.maxBpm`

src/main/java/heronarts/lx/Tempo.java
```diff
@@ -191,16 +194,38 @@ public Tempo(LX lx) {
...
+  }
 
   @Override
   public boolean handleOscMessage(OscMessage message, String[] parts, int index) {
     if (parts[index].equals(PATH_SET_BPM)) {
       if (message.size() > 0) {
-        this.bpm.setValue(message.getFloat());
+        float newBpm = message.getFloat();
+        if (isValidBpm(newBpm)) {
+          this.bpm.setValue(message.getFloat());
```
mcslee: This should be `this.bpm.setValue(newBpm);`

`message.getFloat()` will try to pull ANOTHER argument off the OSC message, after you've already called `message.getFloat()` once.

I might throw an error log message in the case that the BPM is not valid... it's probably good to know when a client is sending invalid data. Understand that it may not always be possible to prevent with 3rd-party OSC software so this change seems sensible in general, but invalid messages are def something to try to get out of the system.

mcslee (changes requested):
Nice - seems sensible, thanks for this. Various super-nitpicky LX-style comments, and just one crucial change, your 2nd call to `message.getFloat()` needs to be `newBpm` to work properly.

Also looks like the email you pushed this change with doesn't match your github account, will just need to sort that out on your commit to be able to sign the CLI.

Other than that happy to merge this once that's resolved - thx!

mcslee:
Sorry just thinking on this more... I realized that when you `setBpmRange` - you don't actually change the valid range of the underlying `BoundedParameter` - it's only changing the range of values accepted by the particular OSC message for setting BPM.

To make this more clear, I think that method and the variables should be explicitly named `setOscBpmRange` - and `minOscBpm`

Because it will still be totally possible via the UI or explicit call for someone to set the `bpm` parameter to a value that is outside of the given range. The range-limiting is only applied on the particular OSC pathway. Thx!

src/main/java/heronarts/lx/Tempo.java
worldveil: i actually am totally in the "write out this. everytime camp". glad to see someone else is too

worldveil: we don't always in TE

src/main/java/heronarts/lx/Tempo.java
worldveil: OH wow. welp that explains a lot of the issues I was having with OSC in LX.....

worldveil:
@mcslee thanks for all the comments here! think I addressed them all, but let me know.

mcslee:
Great - looks good to me... I'm happy to merge this but can you just sort out getting the CLA signed first?

worldveil:
@mcslee I did. the first couple commits are just my laptop before I attached my Github email properly with `git config ...` 

i am fully onboard with the CLA and have signed it, and have made all the commits here. 

if that doesn't work, I'll just start over with a new PR, b/c I don't know how to relabel old commits with an email

mcslee:
Ah yeah sorry about that, GitHub was still flashing a warning at me, but I see now that you had actually already signed it and it's only saying that because of the earlier commit. No problemo - merging now!

PR #68, opened 2022-07-21
Add performance mode to midi surfaces

jkbelcher:
Add performance mode to APC40MkII and MidiFighterTwister.

I used the Session button on the APC to toggle performance mode because in the dark the top-right button is the most convenient to locate, like the escape key.

The MFT target is an enum because I used to have 3 possible targets.  It could be simplified to a boolean.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
@@ -586,6 +586,7 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
       }
       channel.addClipListener(this);
       channel.cueActive.addListener(this);
+      channel.auxActive.addListener(this);
       channel.enabled.addListener(this);
       channel.crossfadeGroup.addListener(this);
       channel.arm.addListener(this);
@@ -612,6 +613,7 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
       }
       this.channel.removeClipListener(this);
       this.channel.cueActive.removeListener(this);
+      this.channel.auxActive.removeListener(this);
       this.channel.enabled.removeListener(this);
       this.channel.crossfadeGroup.removeListener(this);
       this.channel.arm.removeListener(this);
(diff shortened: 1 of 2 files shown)


mcslee:
Oooh, forgot that the PLAY RECORD SESSION buttons hadn't been used and all have LEDs on them!

Now I'm thinking PLAY could be a good toggle for performance mode on/off, and RECORD / SESSION can toggle between cue/aux mode.

I'd been planning to use BANK <- -> to indicate cue vs. aux, but having that up top could be a lot better. Great idea.

jkbelcher:
I started with PLAY for performance mode on/off, but couldn't manage to hit it in the dark without peering at the controller.  SESSION is nice because you can find it by feel.

Seems like DEVICE <- -> should be kept as-is, but BANK <- -> would be useful for changing the target page within the device?

mcslee:
Cool - okay I'm happy to do SESSION for on/off, and I can rationalize this
as then RED button RECORD means AUX mode, matching RED lights for AUX cue.
And GRAY button PLAY means normal cue mode, which is blue which is closer
to gray?? Fuzzy... but close.

Agree BANK buttons will make the most sense to literally use for knob bank
changes. DEVICE selection buttons as is.

I am waffling on whether to change the behavior of the BANK SELECT quad
arrows to work more visually in performance mode, e.g. kind of feels like
up/down wants to browse through devices, but it's also nice how that's
locked to only browsing the pattern list. So I'll probably leave it as-is
to avoid too much overloading.

raldi:
In case either of you (or a visitor from the future) finds it handy, here's a document listing all the APCMk2 buttons and what LX uses them for: https://docs.google.com/document/d/110qgYR_4wtE0gN8K3QZdqU75Xq0W115qe7J6mcgm1So/edit

jkbelcher:
Ah, very nice.  PLAY=cue target and RECORD=aux target.  Better than a single button toggle, because you want to go to a specific one, whereas toggling requires knowing which one you came from.  And SESSION for performance mode on/off, I like it.

jkbelcher:
> I am waffling on whether to change the behavior of the BANK SELECT quad arrows to work more visually in performance mode, e.g. kind of feels like up/down wants to browse through devices, but it's also nice how that's locked to only browsing the pattern list. So I'll probably leave it as-is to avoid too much overloading.

Seems logical to keep the BANK SELECT quad arrows as-is for the selected channel and pattern.  But in performance mode should we do SHIFT+BANK SELECT to navigate the Aux focused channel and Aux pattern list?  Doesn't seem like too big a loss to remove the Jump-to-end function currently available with SHIFT.

jkbelcher:
> In case either of you (or a visitor from the future) finds it handy, here's a document listing all the APCMk2 buttons and what LX uses them for: https://docs.google.com/document/d/110qgYR_4wtE0gN8K3QZdqU75Xq0W115qe7J6mcgm1So/edit

@raldi Nice!

mcslee:
> Seems logical to keep the BANK SELECT quad arrows as-is for the selected channel and pattern. But in performance mode should we do SHIFT+BANK SELECT to navigate the Aux focused channel and Aux pattern list? Doesn't seem like too big a loss to remove the Jump-to-end function currently available with SHIFT.

For this, I just have the BANK SELECT arrows be aware of cue vs. aux in performance mode. No shift button necessary. If you're in CUE mode it navigates the left side, if you're in AUX mode it navigates the right. Same as how the DEVICE <> buttons above know which side you're focused on, as well as the DEVICE knobs. I think that's the right move there. If you're set to AUX mode, no need to use the SHIFT key to navigate around.

jkbelcher:
> For this, I just have the BANK SELECT arrows be aware of cue vs. aux in performance mode. No shift button necessary. If you're in CUE mode it navigates the left side, if you're in AUX mode it navigates the right. Same as how the DEVICE <> buttons above know which side you're focused on, as well as the DEVICE knobs. I think that's the right move there. If you're set to AUX mode, no need to use the SHIFT key to navigate around.

Ah yes, that seems good.  Most of this DEVICE CONTROL button group is now relative to the Cue/Aux selection, except for the buttons previously assigned to custom modes.

mcslee:
FYI finished up the last of the APC40 stuff here (other than knob BANK SELECTION as a future project).
https://github.com/heronarts/LX/commit/ed08129549bfbf9042924d03c438a73b3bbafc0e

I made the CLIP STOP row of buttons control and display channel selection for the *alternative* set of channels in performance mode. So the main track focus row always controls whichever side you're controlling in performance mode. Could be cue, or aux.

The row above, CLIP STOP, is available when *not* in palette mode (where it's used to copy swatches) or clip mode (where it's used to stop clips). In that case, it'll show the focus state and let you focus the side of performance mode that you are *not* focused on.

Figured this was a better tradeoff to ensure that that track select row is always tightly tied to selection of the DEVICE CONTROL knobs, no matter what.

I also preserved bankOn state across dev lock changes, so switching to palette mode doesn't always kick you back to CLIP mode when you leave it.

jkbelcher:
> I made the CLIP STOP row of buttons control and display channel selection for the _alternative_ set of channels in performance mode. So the main track focus row always controls whichever side you're controlling in performance mode. Could be cue, or aux.

That sounds great.

PR #72, opened 2022-07-26
Update MidiFighterTwister for performance mode

jkbelcher:
Uses top button on right side of controller to toggle Aux/Primary.  The red/blue knob colors are a good indicator of state so it seems reasonable to use toggle and keep a button available for future purposes.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
+++ b/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
@@ -20,8 +20,6 @@ package heronarts.lx.midi.surface;
 
 import heronarts.lx.LX;
 import heronarts.lx.LXDeviceComponent;
-import heronarts.lx.color.DiscreteColorParameter;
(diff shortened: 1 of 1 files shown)


src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
```diff
@@ -226,6 +231,13 @@ private void resend() {
       }
     }
 
+    void registerFocusedChannel() {
```
mcslee, line 234: Not worth re-uploading here but toss a `private` on there in a future rev

mcslee (approved):

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: I'll change it on the next edit.  It was copied from the APC40.. I'll let you fix that one since you have a list going.

PR #73, opened 2022-08-10
Use midi sysex to apply lx-friendly config to MFT

jkbelcher:
Tested and working on my MFT.  Have not yet tested with more than one MFT attached to system.

Currently:
-Sets global settings and encoder settings on initialization
-Sends updated encoder settings on device change (currently the only configuration adjusted is Detent aka Bipolar)

Structure is in place to save current MFT config so it can be cleanly re-applied at shutdown.  Need a sysex call/response mechanism to pull current values.

Unfortunately - at the last minute discovered that when the MFT receives an encoder config it disables the display, as the firmware only anticipated receiving encoder updates in combination with a full system config update.  From my initial look at the firmware it appears the only method of re-enabling the display is to send global settings and trigger a reboot, which adds reboot lag plus startup animation lag.  We'll have to decide if the lag is too inconvenient to be worth updating the encoder detent setting on target device changes.  Note I set it up to avoid rebooting if no knob detent settings are changed, so a reboot doesn't happen if you bounce around between patterns with either no bipolar parameters or the same alignment of bipolar parameters.  (In my case - I use a common base class for patterns so the alignment of bipolar parameters is usually the same).

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
+++ b/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
@@ -745,7 +745,6 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
         } else {
           sendControlChange(CHANNEL_ANIMATIONS_AND_BRIGHTNESS, DEVICE_KNOB + i, RGB_ANIMATION_NONE);
           sendControlChange(CHANNEL_ANIMATIONS_AND_BRIGHTNESS, DEVICE_KNOB + i, INDICATOR_ANIMATION_NONE);
-          sendControlChange(CHANNEL_ANIMATIONS_AND_BRIGHTNESS, DEVICE_KNOB + i, INDICATOR_BRIGHTNESS_OFF);
           sendControlChange(CHANNEL_ANIMATIONS_AND_BRIGHTNESS, DEVICE_KNOB + i, INDICATOR_BRIGHTNESS_25);
           if (i <= lxConfig.encoders.length && lxConfig.encoders[i].has_detent.value == CFG_TRUE) {
             sendControlChange(CHANNEL_ROTARY_ENCODER, DEVICE_KNOB+i, 63);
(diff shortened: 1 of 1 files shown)


mcslee:
Nice - looks good. I have a bit of stylistic feedback on it - could clean up a lot of settings calls by registering those in a dictionary or list. Hopefully later today will have some time to merge and fire back some cleanup your way.

Man that's a bummer about having to send the global reset if encoders change. Not much we can do about that without a firmware update I guess. Unfortunate. I wonder if that knocks the MIDI I/O offline. The only thing I'd be concerned about there is if rapid pattern focus changes happen in succession, is there a possibility that the later update gets lost if the sysex is sent while the device is still busy rebooting. Or does the MIDI input still get queued.

jkbelcher:
For sure, would appreciate some stylistic cleanup here.

Looks like the midi input still goes through after rapid pattern changes.  I did 30 quick changes with the keyboard to test.  It does accumulate a delay while it's rebooting 30 times... takes a while to catch up from that.  I don't know if it's bad for the hardware to be flashing so frequently.

Could have a surface setting to disable bipolar lights in case the reboot is a hassle for some situations.

Another idea: run this as default, but compile a custom firmware version for serious LX users that allows the bipolar setting to be modified with a normal midi command.  We could detect the custom firmware version from the identifier sysex and use the extra midi function when available.

mcslee:
Good stuff - I don't actually mind the global update doing the animation on the control surface when the knob styles change. It's not too slow and it serves reasonably well to let you know that the device focus changed.

Here are cleanups, mostly I store the settings in a hash map and loop over that rather than lots of duplicated per-setting code. Also renamed the inner classes to be more simple. In general, don't need to re-add a prefix like MFTWhatever if you're already in a nested inner class of MidiFighterTwister. the full class name is already MidiFighterTwister.Config.Encoder.Setting etc.

https://github.com/heronarts/LX/commit/3d574f0b3aa7e689442f2ea0b8ebf09b6582a9de

PR #74, opened 2022-08-21
Allow easy crossfader sync from DJM900nxs2

jkbelcher:
Proposed auto-integration.  Discussion sent by email.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
@@ -724,9 +724,14 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
 
   }
 
+  public final BooleanParameter disableXF =
+    new BooleanParameter("Disable Crossfader", false)
+    .setDescription("Prevents interference with crossfader control from another device");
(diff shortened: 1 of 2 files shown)


mcslee (commented):
I'm on board with this, will do a few post-cleanups. Only part I'm not cool with is reaching into the APC40 settings and changing them. I think that should be a step done-by-the-user manually, otherwise I think it's just as likely to cause confusion as to fix it (e.g. you had your APC40 set up, you clicked the Pioneer options a couple times just to see what they were, and now your APC40 crossfader isn't working but you don't know why and didn't notice it was changed).

mcslee:
Touched up here: https://github.com/heronarts/LX/commit/c0b8dcce1508261a168ac935b517da4452ede92e

jkbelcher:
> (e.g. you had your APC40 set up, you clicked the Pioneer options a couple times just to see what they were, and now your APC40 crossfader isn't working but you don't know why and didn't notice it was changed

Good call.  Auto-toggling would add confusion unless there was a central option to choose crossfader source, channel faders source, etc.  But that's doesn't really work either to assume only one can be a source.  Because there are motorized faders out there... :).  This looks great for now.

PR #76, opened 2022-10-09
Add inverse trig functions to LXF parsing

jkbelcher:
The trig functions in LXFs are super useful. Already had a case where I needed inverse trig to calculate angle from opposite / adjacent.


mcslee:
I'm scared atan is going to lead to wanting the two-argument atan2... and oh god then possibly having to start parsing argument lists... holding my breath!

jkbelcher:
Look I know we're at capacity but atan2 says all his friends are inside and wants to know if he can get in the club, too.

Not needed *quite* yet...

PR #77, opened 2022-10-27
Add modulus to simple expression operators

jkbelcher:


mcslee:
Updated this a bit:
https://github.com/heronarts/LX/commit/34999453684696d07ae7d4b8b2c87c874d785bf7

Doing proper operator precedence, so that it matches C where these operators have equal precedence and associate left-to-right:
4 * 5 % 9 => will evaluate as (4 * 5) % 9
4 % 5 * 9 => will evaluate as (4 % 5) * 9

jkbelcher:
Thanks!

PR #78, opened 2022-11-28
Update a few OSC parameters to use absolute values

jkbelcher:
First pass doing some OSC cleanup.

### DIFF
--- a/src/main/java/heronarts/lx/LXEngine.java
+++ b/src/main/java/heronarts/lx/LXEngine.java
@@ -93,9 +93,10 @@ public class LXEngine extends LXComponent implements LXOscComponent, LXModulatio
 
   public final Output output;
 
-  public final BoundedParameter framesPerSecond =
+  public final BoundedParameter framesPerSecond = (BoundedParameter)
     new BoundedParameter("FPS", 60, 1, 300)
     .setMappable(false)
+    .setOscMode(BoundedParameter.OscMode.ABSOLUTE)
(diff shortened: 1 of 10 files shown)


mcslee:
In Hawaii at the moment so a bit slow. I'm inclined to go more under-the-hood here, and have any parameter which has units specified as `SECONDS` or `MILLISECONDS` automatically use absolute osc mode. This could be overridden manually to set the parameter back to normalized mode, but it seems like in 99% of cases time values are going to want to be absolute.

If something like a transition time is *not* absolute, it should probably actually be using something like `PERCENT` for the units.

I'll take a look at both this, and changing the handling of `DiscreteParameter` when I'm settled back from travels.

jkbelcher:
Yep, going off of units=`SECONDS` did cross my mind.  Seems like a fine place to do it.

The non-seconds parameters in this PR are `framesPerSecond` and `gamma`.

mcslee:
Subsumed by structural changes to make all DiscreteParameters ABSOLUTE mode, and for types like SECONDS/MILLISECONDS to default to that type.

https://github.com/heronarts/LX/commit/b09939cb7c65c62da9b1bd5f761ba2cd67273db0
https://github.com/heronarts/LX/commit/07ac6496263849587646d5e2035e3eb2d9f63a97
https://github.com/heronarts/LX/commit/e2611a8e0a30857a1172af716bbad75121542b04

PR #80, opened 2023-03-30
APC bug fixes

jkbelcher:
`Device.enabled` is now included in `LXDeviceComponent.defaultRemoteControls`.  In the absence of custom remote controls this caused an IllegalStateException in `APC40Mk2.registerDeviceKnobs():419` when the surface tried to add a second listener to `enabled`.  `Enabled` is special and is already listened to by DeviceListener.

This PR adds special handling for `enabled` and tweaks the Dev On/Off light to stay in sync when compositeMode is changed.


mcslee:
Nice, APC40 mk1 might need the same fix too, but not overly worried about it and don't have one physically here with me...

PR #81, opened 2023-04-02
Clean up MFT behavior for stray encoder readings

jkbelcher:
Found two use cases where MFT encoder sends a value outside of previously expected range:
1. Knob is rotated during config sysex / reboot cycle.
2. After two dusty events last summer, parameters were occasionally jumpy on MFT knob turn even when configured correctly (sending relative values).  Current theory is dust in the internals causes a jump in the reading which is interpreted as an extremely fast increment/decrement.

Previous behavior was values fall through and get interpreted as absolute.  This causes a harsh parameter jump and visual glitch which is undesired.

This PR constrains encoder inputs to expected range to avoid unwanted behavior on a dusty system.  If it appears after further field use that some of these values are on the wrong side of the expected range, then will have to adjust this code to discard the values completely.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
+++ b/src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
@@ -764,6 +764,7 @@ public class MidiFighterTwister extends LXMidiSurface implements LXMidiSurface.B
       }
     }
 
+    @SuppressWarnings("unused")
     private void onKnob(int index, double normalized) {
       if (this.knobs[index] != null) {
         this.knobs[index].setNormalized(normalized);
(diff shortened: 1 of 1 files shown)


src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
```diff
@@ -764,6 +764,7 @@ public void onParameterChanged(LXParameter parameter) {
       }
     }
 
+    @SuppressWarnings("unused")
```
mcslee: I'd just delete this code, it's dead now

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
```diff
@@ -1009,15 +1010,19 @@ public void controlChangeReceived(MidiControlChange cc) {
       case CHANNEL_ROTARY_ENCODER:
         if (number >= DEVICE_KNOB && number <= DEVICE_KNOB_MAX) {
           int iKnob = number - DEVICE_KNOB;
+          if (value < KNOB_DECREMENT_VERYFAST || value > KNOB_INCREMENT_VERYFAST) {
```
mcslee: I would leave this whole block in the `else` like it was before. Handle all the valid values first, then have this error block.

Doing the "is-it-an-error" conditional check before the valid values seems a bit more confusing to read through.

mcslee (changes requested):
Minor nitpicks. Def something to keep an eye on in practice... if it happens a whole lot may need to get some kind of feedback up in to the UI for the user to be made aware why the knob is doing nothing. But that'll be a pain in the ass. This is fine for now.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: Done.

src/main/java/heronarts/lx/midi/surface/MidiFighterTwister.java
jkbelcher: Sounds good.  Fixed.

PR #86, opened 2023-04-15
Add DJM-V10 mixer

jkbelcher:
Had the chance to borrow one of these.  Sound quality is excellent btw.

Adds the same smart crossfader functionality as the DJM900nxs2.


mcslee:
Nice. In theory seems like there could be some code-sharing between this and the 900, but unlikely that these will change enough for that to matter a whole lot.

PR #88, opened 2023-05-30
Fix Maven build out of box by avoiding error on JAVA_HOME

jkbelcher:
By default the dev toolchain produces an error running maven clean package within eclipse, on OSX/Ventura/apple arch:

`Error while generating Javadoc: Unable to find javadoc command: The environment variable JAVA_HOME is not correctly set.
`

JAVA_HOME is not set by the installers, but this addition to pom solves the problem.  Worth inclusion to make things build out of the box?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -75,15 +75,6 @@
 	<build>
 		<plugins>
 			<plugin>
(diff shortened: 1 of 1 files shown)


mcslee:
Wow bizarre, never seen that before!

PR #89, opened 2023-06-01
Return normalized value of zero for single-int DiscreteParameter

jkbelcher:
A DiscreteParameter can exist with a range of 1, ie a single value.  Calling getNormalized() on this parameter was returning NaN from the divide by zero.


mcslee:
Yikes, thx for catching.

PR #91, opened 2023-06-16
Fix object comparison for default value in ObjectParameter constructor

jkbelcher:


mcslee (changes requested):
Sorry just realized to be fully safe here, we should check for two nulls. It's obviously a terrible idea to put nulls in here, but also a bad idea to `NullPointerException` if someone does.

mcslee:
Fixed here: https://github.com/heronarts/LX/commit/4edf4ecd18c110e2e590b679c3626b1b35bc7d67

jkbelcher:
So `value` is not null as checked by the preceding lines.  I was only thinking of the immediate use case of Strings, where string.equals(null) should return false.  Are you concerned some object will throw an exception on .equals(null)?

mcslee:
Oh good point forgot null is checked above. In any case Objects.equals that
I pushed in there will be fine

Le sam. 17 juin 2023 à 06:01, Justin Belcher ***@***.***> a
écrit :

PR #92, opened 2023-06-29
Release buffer from LXLayeredComponent

jkbelcher:
Unsure how important this is... seems like a good practice for garbage collector.

### DIFF
--- a/src/main/java/heronarts/lx/LXLayeredComponent.java
+++ b/src/main/java/heronarts/lx/LXLayeredComponent.java
@@ -178,6 +178,7 @@ public abstract class LXLayeredComponent extends LXModelComponent implements LXL
       layer.dispose();
     }
     this.mutableLayers.clear();
+    this.buffer = null;
     super.dispose();
   }
 


mcslee:
This shouldn't be necessary. If the component is properly disposed and this was the only reference to that buffer, it will be garbage collected. If that's not happening, then there's a deeper issue in the dispose cycle - but I don't think we need to be setting fields to `null` in `dispose()` calls.

jkbelcher:
This is a reference to renderBuffer or blendBuffer passed by LXChannel to a pattern or effect.  So if you run 20 patterns and then delete them, and then delete the channel, there will be 21 references to the LXBuffer.  Can the GC identify the whole group of objects as unreachable when they have references to each other?  And the base LXComponent class has a reference to the LX instance.

mcslee:
Yeah, the GC should catch that. When the pattern/effect is disposed, all references to it should go away, and it should end up unreachable. This will result in its reference to the buffer going away, and then when the LXChannel gets disposed, its reference will also go away, and all should become unreachable.

> And the base LXComponent class has a reference to the LX instance.

That's fine, since that's a reference to LX, not *from* LX. The issue here would be if something else was retaining references *to* the post-disposed channel/pattern/effect/buffer.

If they are, that's a bug that needs to be fixed at its source.

PR #100, opened 2023-10-18
DMX modulators

jkbelcher:
It sure was nice to have the LXDmxEngine scaffolding in place leading into this...

Here's a set of basic DMX modulators to handle some common functions:

- Dmx16bitModulator: consume two channels of DMX for high resolution
- DmxColorModulator: receive a color with three channels and route it to an entry in the global palette
- DmxRangeModulator: extract a limited range of values within a channel, for combining multiple features into one channel.
- DmxSplitModulator: like creating two DmxRangeModulators, but a quick and easy way to split a channel into two functions.
- DmxGridModulator: route a sequential set of channels to a 2D array.

Added a category for DMX.  It feels like the original DmxModulator should be renamed but didn't take any guesses at that yet.

Range and Split both include boolean "active" parameters that are true when the received value is within range.  It works well to map Active to an effect.enabled parameter and map the range output to the effect amount/speed/etc.

UIs are ready and will get cleaned up when time permits.


src/main/java/heronarts/lx/dmx/Dmx16bitModulator.java
```diff
@@ -0,0 +1,52 @@
...
+@LXCategory(LXCategory.DMX)
+public class Dmx16bitModulator extends DmxModulator {
+
+  public Dmx16bitModulator() {
+    this("DMX 16-bit");
+  }
+
+  public Dmx16bitModulator(String label) {
+    super(label);
+    this.channel.setRange(0, LXDmxEngine.MAX_CHANNEL - 1);
```
mcslee: This is fine as long as the two bytes never cross a universe boundary, which would seem like a really dumb thing to do. But just calling it out here.

src/main/java/heronarts/lx/dmx/DmxColorModulator.java
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Extracts a color from three DMX channels starting at a given address.
+ */
+@LXModulator.Global("DMX Color")
+@LXModulator.Device("DMX Color")
+@LXCategory(LXCategory.DMX)
+public class DmxColorModulator extends DmxModulator {
+
+  public enum ColorPosition {
```
mcslee: This feels ugly, I think I'd rather have two parameters here, one explicit BooleanParameter which is on/off for whether "push-to-palette" is enabled. And I'll note that this behavior feels like it sort of breaks the modulator self-containment contract. So I'd want to keep that default false, but I can see how it's very useful, so I'm fine to make it an option.

Then the destination can be a simple `DiscreteParameter` with range `(0, LXSwatch.MAX_COLORS)`. No need for a whole new enum here.

I think this modulator should expose its color value in its UI, so you could always achieve a similar result by making parameter mappings to control hue/sat/brightness.

src/main/java/heronarts/lx/dmx/DmxGridModulator.java
```diff
@@ -0,0 +1,117 @@
...
+import heronarts.lx.parameter.DiscreteParameter;
+import heronarts.lx.parameter.LXParameter;
+
+/**
+ * Maps a sequential set of DMX channels onto a 2D grid of size [rows x columns]
+ */
+@LXModulator.Global("DMX Grid")
+@LXModulator.Device("DMX Grid")
+@LXCategory(LXCategory.DMX)
+public class DmxGridModulator extends DmxModulator {
```
mcslee: I'm not sure if this should really be called a Grid rather than just a generic range. The fact that there are rows and columns is not really relevant to its operation... the only thing the algorithm cares about is total number of values being averaged.

There could still be a helper for row/col based indexing, but that's really just a utility helper-method. The function of the modulator itself is to average a range.

I think "Range" as a name would probably intuitively correspond to this behavior. Or if we have a DMX section, than really I think this could even just be called "Average" - that's actually the clearest really.

Is there a particular use case you guys have where this being a grid is really relevant?

src/main/java/heronarts/lx/dmx/DmxGridModulator.java
```diff
@@ -0,0 +1,117 @@
...
+          // Grid did not fit within ArtNet data
+          break;
+        }
+      }
+    }
+
+    return sum / (double)resolution / 255.;
+  }
+
+  public int getValue(int row, int column) {
```
mcslee: This helper could still exist somewhere, with grid size passed as arguments. But I'd like to better understand what this is being used for. Seems like it could be a helper in grid-based client code, rather than in this modulator API itself.

src/main/java/heronarts/lx/dmx/DmxRangeModulator.java
```diff
@@ -0,0 +1,107 @@
...
+import heronarts.lx.parameter.BooleanParameter;
+import heronarts.lx.parameter.DiscreteParameter;
+import heronarts.lx.parameter.LXParameter;
+
+/**
+ * Modulator receiving a range from [min] to [max] within a DMX channel.
+ */
+@LXModulator.Global("DMX Range")
+@LXModulator.Device("DMX Range")
+@LXCategory(LXCategory.DMX)
```
mcslee: Not sure that Range is an intuitive name for this. Something like Conditional or Trigger maybe seems more appropriate, since that's what it is outputting.

src/main/java/heronarts/lx/dmx/DmxRangeModulator.java
```diff
@@ -0,0 +1,107 @@
...
+
+  public final DiscreteParameter min =
+    new DiscreteParameter("Min", 0, 256)
+    .setDescription("Minimum input value for range");
+
+  public final DiscreteParameter max =
+    new DiscreteParameter("Max", 255, 0, 256)
+    .setDescription("Maximum input value for range");
+
+  public final BooleanParameter active =
```
mcslee: My typical convention for boolean modulators is to call this variable `out`

src/main/java/heronarts/lx/dmx/DmxRangeModulator.java
mcslee: Also `implements LXTriggerSource`

src/main/java/heronarts/lx/dmx/DmxRangeModulator.java
```diff
@@ -0,0 +1,107 @@
...
+      }
+      return ((double) dmx - min) / (max - min);
+    } else {
+      this.active.setValue(false);
+      return 0;
+    }
+  }
+
+  protected int getDmxValuei(int universe, int channel) {
+    return this.lx.engine.dmx.getByte(universe, channel) & 0xff;
```
mcslee: Just use `this.lx.engine.dmx.getValuei(universe, channel)` - this helper already exists

src/main/java/heronarts/lx/dmx/DmxRangeModulator.java
```diff
@@ -0,0 +1,107 @@
...
+  }
```
mcslee:
```
  @Override
  public BooleanParameter getTriggerSource() {
    return this.out;
  }
```

src/main/java/heronarts/lx/dmx/DmxSplitModulator.java
```diff
@@ -0,0 +1,109 @@
...
+ *  0        Output1 = 0    Output2 = 0
+ *  1-127    Output1 = 0-1  Output2 = 0
+ *  128-255  Output1 = 0    Output2 = 0-1
+ *
+ * @author Justin K. Belcher
+ */
+@LXModulator.Global("DMX Split")
+@LXModulator.Device("DMX Split")
+@LXCategory(LXCategory.DMX)
+public class DmxSplitModulator extends DmxModulator {
```
mcslee: This is pretty strange, is this a common hardware behavior? What does this?

This whole thing is kind of giving me the heebeegeebees... also wondering if the utility math stuff like `Operator` modulator can't just be used to implement this behavior.

Is this a common thing?

src/main/java/heronarts/lx/dmx/DmxSplitModulator.java
```diff
@@ -0,0 +1,109 @@
...
+
+
+  public final BooleanParameter isZero =
```
mcslee: Oof weird, why is this a separate case from output1 and output2 both just being zero? Is there some hardware where this is a semantically meaningful distinction?

src/main/java/heronarts/lx/dmx/DmxSplitModulator.java
```diff
@@ -0,0 +1,109 @@
...
+
+    if (dmx == 0) {
+      this.isZero.setValue(true);
+      this.active1.setValue(false);
+      this.active2.setValue(false);
+      this.output1.setValue(0);
+      this.output2.setValue(0);
+      return 0;
+    } else if (dmx < 128) {
+      double value = (dmx - 1) / 126.;
```
mcslee: Checking if this is right. Is DMX value "1" supposed to produce an output of "0"?

Seems like this should just be `dmx / 127.`?

Otherwise this seems quite weird that the first "half" is 127 options, the second "half" is 128 options and 0 is special?

mcslee (changes requested):
Got some high level questions on the use case for a few of these, I'd like to think a bit about what the UI for them looks like, whether all are necessary and/or how to make naming a bit more clear.

Various comments/questions inline.

src/main/java/heronarts/lx/dmx/DmxModulator.java
```diff
@@ -26,7 +26,7 @@
 
 @LXModulator.Global("DMX")
```
mcslee, line 39: With a new DMX category, it seems like the UI name for this could just be "Channel", and we could drop the "DMX" prefix from the other modulator device names?

src/main/java/heronarts/lx/dmx/Dmx16bitModulator.java
jkbelcher: Two universes would be different packets, rights?  So it seems reasonable to force a two-byte data point into a single universe as we're doing here, to prevent the two bytes from being sourced from different frames.

src/main/java/heronarts/lx/dmx/DmxGridModulator.java
jkbelcher: Ah sorry this one wasn't clear.  The purpose is not to extract an average... I was thinking modulators internally contain a single value and the average was more useful than zero.  The main purpose here was to expose the getValue(row, column) method.

We're using it to receive a low-resolution grid from a grandMA console, overlayed onto the fixtures of the car. Allowing the console to map a swipe effect across their fixtures outside of the car and within the car itself.  Demo vid:
https://www.youtube.com/watch?v=sPlDItBaxd0

src/main/java/heronarts/lx/dmx/DmxRangeModulator.java
jkbelcher: Copy/paste error from the other repo before that method was available.  Fixed.

src/main/java/heronarts/lx/dmx/DmxGridModulator.java
mcslee: The use case is cool but I'm not really seeing why there is a modulator here. Shouldn't these just be parameters and a helper method on your Dmx Grid pattern implementation? Is your Dmx Grid pattern looking for a global modulator of the DmxGrid type? That feels hacky/messy.

This one feels like it ought to just live in pattern land as utility code, not as a user-facing modulator. The grid settings in the modulator don't do anything for a user, only for a programmer that is coding against the helper method API. And I don't think you need all this overhead of the modulator separately caching its own `values[][]` array and recomputing it on every draw loop. Feels like your pattern can just ask the DmxEngine directly for the values it wants when it needs them, and there's no extra processing that way.

jkbelcher: Yep, tunnel vision on routing everything through modulators.  Removed.

src/main/java/heronarts/lx/dmx/DmxGridModulator.java
jkbelcher: Removed, per above.

src/main/java/heronarts/lx/dmx/DmxSplitModulator.java
jkbelcher: In retrospect this was built prior to the DmxRangeModulator, and is really just two DmxRangeModulators without being as flexible.  Removed!

src/main/java/heronarts/lx/dmx/DmxSplitModulator.java
jkbelcher: removed, per above

src/main/java/heronarts/lx/dmx/DmxSplitModulator.java
jkbelcher: removed this modulator, per above

src/main/java/heronarts/lx/dmx/DmxModulator.java
jkbelcher: For sure.  Seems like the names of this modulator and Dmx16bitModulator should match.  Like "1-Channel and 2-Channel" or "8-bit and 16-bit" or "Basic/Simple and High-Resolution".

The DMX prefix isn't needed in the Add Modulator menu when grouped by category, but should we leave it for the default modulator name?  Otherwise, a modulator of name "Channel" in the modulators instance list seems a little vague, could be confused with something related to a LX Mixer Channel.

src/main/java/heronarts/lx/dmx/DmxColorModulator.java
jkbelcher: > This feels ugly, I think I'd rather have two parameters here, one explicit BooleanParameter which is on/off for whether "push-to-palette" is enabled.

Agreed, cleaner without the enum.  Removed.

> And I'll note that this behavior feels like it sort of breaks the modulator self-containment contract.

True.  Could there be a way to map a ColorParameter to a LXDynamicColor?  So the output of this modulator would be a color, and you could click Map on this modulator and click the target color in a swatch.

src/main/java/heronarts/lx/dmx/DmxRangeModulator.java
jkbelcher: > Also implements LXTriggerSource

Yes!  Thanks.

jkbelcher: > Not sure that Range is an intuitive name for this. Something like Conditional or Trigger maybe seems more appropriate, since that's what it is outputting.

My primary intention was to map the range of input values to a normalized output.  The boolean out is helpful too, but the names Conditional or Trigger might hide the normalized function.  Other ideas?

This is common with DMX fixtures, to divide one channel into multiple alternative ranges.  Example:
image

And an example where some ranges within a DMX channel are boolean and some are normalized:
image
...In the case of 64-95 for strobe effect slow to fast, for example I would map the boolean `out` of this modulator to a LX `StrobeEffect.enabled` parameter, and map the normalized range to the `StrobeEffect.speed` parameter.

src/main/java/heronarts/lx/dmx/DmxRangeModulator.java
jkbelcher: Done, thx.

src/main/java/heronarts/lx/dmx/DmxModulator.java
jkbelcher: Removed DMX prefix from Add Modulator menu.

jkbelcher:
Ready for another review.

jkbelcher:
I could also just merge the original DmxModulator, 16-bit, and Range.  That would make a flexible modulator that would cover a lot of use cases.

mcslee:
>  I could also just merge the original DmxModulator, 16-bit, and Range. That would make a flexible modulator that would cover a lot of use cases.

Yeah I agree, I think that's the best approach actually. Can have an `EnumParameter` in the `DmxModulator` for mode, 8-bit/16-bit/Range. When mode is set to 16-bit the max channel is restricted to 510 rather than 511. When mode is range then the min/max parameters are also used.

I can put together the UI updates for this when it's sorted.

jkbelcher:
Merged those modulators, plus a couple add-ons:

- I pulled the max channel restriction based on number of bytes into an abstract base class.  Common functionality for now, but also: Can imagine this being useful down the road in a drag-and-drop "construct a DMX profile" scenario, where a list of dmx modulators and their byte size allows an auto-calc of the channel assignments.
- Removed LXNormalizedParameter from DmxColorModulator.  Now the modulator value is the color wrapped in a double.  Would be a nice feature if this was mappable to other color parameters... maybe we could add a LXMappableColor interface?

Happy to revise or feel free to take if from here and tune it up as needed.

mcslee:
Great, all set here. I did a cleanup pass, nothing too major:
https://github.com/heronarts/LX/commit/3844ea4866761decfca9ee0aeafbed55c4241044
https://github.com/heronarts/LX/commit/f715cf5e77a4a582334c6e9e0fb8104665a5fda0

Updated UIs for both of these.
Screen Shot 2023-11-02 at 3 57 35 PM

PR #102, opened 2023-11-08
Add public accessor for AHDSREnvelope.stage

jkbelcher:
I'm using an AHDSREnvelope from code.  It seems like I should be able to check the current stage with a `getStage()` method.  It would also be nice to listen for stage changes rather than poll them every frame.  Would you consider adding a Listener class to this modulator with a `stageChanged()` event?

### DIFF
--- a/src/main/java/heronarts/lx/modulator/AHDSREnvelope.java
+++ b/src/main/java/heronarts/lx/modulator/AHDSREnvelope.java
@@ -284,6 +284,13 @@ public class AHDSREnvelope extends LXModulator implements LXNormalizedParameter 
     }
   }
 
+  /**
+   * Current stage of the envelope.
+   */
+  public Stage getStage() {
+    return this.stage;
+  }
+
   @Override
   public LXNormalizedParameter setNormalized(double value) {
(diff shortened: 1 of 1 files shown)


mcslee:
I guess I'm not opposed to a listener - just curious, what is the use case? What are you doing differently on a stage-by-stage basis?

mcslee:
Here, how's this:
https://github.com/heronarts/LX/commit/aa1dced4d41f69d6c9d6474537a1a91878af2a5f

Rather than dealing with an array of listeners, just using an anonymous subclass pattern here, you can do:

```java
AHDSREnvelope env = new AHDSREnvelope(...) {
  protected void onStageChanged(AHDSREnvelope.Stage stage) {
    // do your thing
  }
};
```

jkbelcher:
Thanks, the anonymous subclass works great here.

The use case is a collection of objects with parallel lifetimes, each with its own envelope defining the lifecycle.  Exact uses still in flux but have included:
1) Is lifecycle over?  getLevel() == 0.0 could mean it was created this frame, or it is past the decay stage.  Can't tell the difference.
2) Trying to log a timestamp at the beginning of various phases, particularly sustain and decay, for triggering events after a certain amount of delay.

PR #104, opened 2024-03-27
Add setting to disable device knob mappings to focused device on APC40mkII

jkbelcher:
APC40mkII device knobs have become redundant when used alongside a MidiFighterTwister.  This adds a setting to bypass the focused device map and free up the knobs to be used for generic mappings.

The primary change here separates the DeviceListener registration from the midi surface registration so it can be toggled independently.

### DIFF
--- a/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APC40Mk2.java
@@ -267,7 +267,9 @@ public class APC40Mk2 extends LXMidiSurface implements LXMidiSurface.Bidirection
 
   private void setAux(boolean isAux) {
     this.isAux = isAux;
-    this.deviceListener.focusedDevice.setAux(isAux);
+    if (isDeviceControl()) {
+      this.deviceListener.focusedDevice.setAux(isAux);
+    }
     this.lx.engine.performanceMode.setValue(true);
     sendPerformanceLights();
     sendCueLights();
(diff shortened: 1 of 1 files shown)


mcslee:
Good stuff - minor cleanup here:
https://github.com/heronarts/LX/commit/afe9840e8d890b479780393c22cab436c0518ff3

Just re-using some knob initialization code that repeats a few spots, and pulling `isDeviceControl()` outside of the loop

mcslee:
FYI: https://github.com/heronarts/LX/commit/b3741894b5ad7ae1d3cefb86633993e8a6854518

I think the BANK SELECT LEFT / RIGHT should still be changing the focused channel even if DEVICE CONTROL is off. Logically it's in a different part of the controller and feels like it's navigation.

jkbelcher:
> FYI: [b374189](https://github.com/heronarts/LX/commit/b3741894b5ad7ae1d3cefb86633993e8a6854518)
> 
> I think the BANK SELECT LEFT / RIGHT should still be changing the focused channel even if DEVICE CONTROL is off. Logically it's in a different part of the controller and feels like it's navigation.

Definitely.  Thanks.

mcslee:
FYI - added echo of input CC parameters to the Device Control knobs when not in Device Control mode. Otherwise these just get stuck with the LED pointers upwards when you turn them.

Update to have them behave like the top row of custom channel knobs:
https://github.com/heronarts/LX/commit/b7644cc00e9f3fa7a4e70cb015452eace229af68

PR #106, opened 2024-04-02
APCmini Mk2

jkbelcher:
Functionality is all the same as the Mk1.  I split this into two commits so you can see the changes from the Mk1 in the second commit.  Changes include:

- LED color & behavior use new midi channels and values.
- Added a masterFaderEnabled setting, same as the APC40mk2
- Pulled a few common functions into methods, mainly because I am mucking with them in an experimental implementation of this surface.


jkbelcher:
I think we could consider different buttons for modifying the Channel Button behavior.  Currently it is Shift+[Fader Ctrl] buttons.  It could instead be the [Solo/Mute/RedArm/Select] buttons on the right side, and the current selection could stay lit.  That would be slightly more discoverable and those buttons are otherwise unused.

mcslee:
Thanks for this! FYI I didn't look very closely. I may review this after RHres stuff, I just bought one of these but probably won't have time to really dig into it until after.

PR #107, opened 2024-04-04
Remove unique encoder click behavior for DiscreteParameter

jkbelcher:
The click-to-increment DiscreteParameters was a nice idea but in practice I've found it to be annoying.  As a user you're expecting a certain behavior from the click (I'm expecting Reset) and it would be more convenient if the behavior was the same for DiscreteParameters.


mcslee:
Yup I'm agreed, looks like the other knobs, consistency feels best.

Issue #105, opened 2024-04-02
MidiFighterTwister generic midi mappings is broken by relative encoders


jkbelcher:
When the MidiFighterTwister is used as a generic midi surface with user-created midi mappings, the encoders are still sending relative position changes which are midi values (61, 62, 63, 65, 66, or 67).  The generic midi mappings are not aware of the relative encoder setting and are interpreting these as absolute midi values.  Turning a mapped knob sets the target parameter to a normalized value of (usually) 63/127 or 65/127.

jkbelcher:
Also: an encoder click mapped to a BooleanParameter uses absolute values which effectively treats the target as a Momentary (encoder down -> parameter on.  encoder up -> parameter off) even if it is Toggle mode.

mcslee:
Yeah... it's vaguely on my list, but not very high, to auto-detect these MIDI controllers that use relative MIDI values. I'm not 100% sure what best practice is on this other than just "noticing" during the mapping phase that the knob isn't sending any other values besides the relative increment/decrement values.

The clicks that use CC with value 0/127 rather than MIDI notes are kind of annoying to deal with without special-casing.

Is your use case that you want to mix-and-match control surface and manual mapping behavior on an MFT? Or just don't want to have to deal with re-configuring the MFT via its utility to work in a more sensible way for generic mapping? (e.g. absolute CC values on the encoders and MIDI notes for the pushes).

jkbelcher:
Not my use case.  This was passed along to me by @bbulkow.  I think someone wanted to use the MFT for generic mapping without applying different settings than we've pushed to it.

jkbelcher:
Given that LXMidiMapping.Note has Mode and DiscreteMode parameters, it seems comparable for LXMidiMapping.ControlChange to have an Absolute/Relative parameter.  Thinking an enum since it looks like there are multiple types of relative encoders out there.

Agree the CC 0/127 seems like a special case.  Although doesn't seem to be a problem to accommodate it in this first version of a CC Relative mode.

mcslee:
Okay gotcha - yeah I'll eventually get to these options, but for the time being my best recommendation is "put the MFT config back onto a more basic template."

PR #108, opened 2024-04-22
Apply settings friendly to generic mapping on MFT disable/shutdown

jkbelcher:
Per recommendation [here](https://github.com/heronarts/LX/issues/105#issuecomment-2032288004), sets the MFT config to a state that plays nice with generic midi mappings.  Also sets background lights to green.


mcslee:
Cool, makes sense, this is friendlier behavior until pull functionality is available (which realistically might be never).

PR #109, opened 2024-05-03
Allow reset to new default value in StringParameter

jkbelcher:
StringParameter had no mechanism to modify the default value after the parameter was created.  Adding a `reset(new default)` method to match other parameter types.


jkbelcher:
Added typo correction for existing Exception message.

PR #110, opened 2024-05-08
Add channel banks and fader pickup to APCminiMk2

jkbelcher:
Added utility class `MixerSurface` which could be used by any midi surface to add channel banking.  Trying it out first on the APCminiMk2.  Some users are interested in running two of these side-by-side to control 16 channels.  Moderately tested.

I threw in a hook for hiding channels.  Have heard users with a single controller lament how quickly the midi controller channels are used up when using groups.  Figured this was a good place to bake that in for future-proofing.

Added `LXPickupParameter` which wraps a target parameter to add a pickup mode when mapped from a one-directional midi control.  This one was a bit rushed, edits welcome.

### DIFF
--- a/src/main/java/heronarts/lx/utils/LXUtils.java
+++ b/src/main/java/heronarts/lx/utils/LXUtils.java
@@ -499,4 +499,17 @@ public class LXUtils {
   public static boolean isEmpty(String s) {
     return s == null || s.trim().isEmpty();
   }
+
+  /**
+   * Compare two objects either of which may be null.
+   */
+  public static boolean equals(Object obj1, Object obj2) {
+    if (obj1 == null) {
+      if (obj2 == null) {
+        return true;
+      }
+      return false;
+    }
+    return obj1.equals(obj2);
+  }
 }
--- a/src/main/java/heronarts/lx/midi/surface/APCminiMk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APCminiMk2.java
@@ -444,7 +444,7 @@ public class APCminiMk2 extends LXMidiSurface implements LXMidiSurface.Bidirecti
 
   }
 
-  private final MixerSurface mixerSurface;
+  protected final MixerSurface mixerSurface;
 
   private final MixerSurface.Listener mixerSurfaceListener = new MixerSurface.Listener() {
     @Override
--- a/src/main/java/heronarts/lx/midi/surface/APCminiMk2.java
+++ b/src/main/java/heronarts/lx/midi/surface/APCminiMk2.java
@@ -36,12 +36,13 @@ import heronarts.lx.mixer.LXBus;
 import heronarts.lx.mixer.LXChannel;
 import heronarts.lx.mixer.LXAbstractChannel;
 import heronarts.lx.mixer.LXGroup;
-import heronarts.lx.mixer.LXMixerEngine;
 import heronarts.lx.parameter.BooleanParameter;
 import heronarts.lx.parameter.DiscreteParameter;
+import heronarts.lx.parameter.EnumParameter;
 import heronarts.lx.parameter.LXListenableNormalizedParameter;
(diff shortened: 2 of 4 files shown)


mcslee:
Just FYI I'm gonna pull this into my own branch and do a lot of surgery on it... will keep you posted.

src/main/java/heronarts/lx/utils/LXUtils.java
```diff
@@ -499,4 +499,17 @@ public float get(double basis) {
   public static boolean isEmpty(String s) {
     return s == null || s.trim().isEmpty();
   }
+
+  /**
+   * Compare two objects either of which may be null.
+   */
+  public static boolean equals(Object obj1, Object obj2) {
```
mcslee, line 506: FYI this is `Objects.equals()` in standard Java
https://docs.oracle.com/javase/8/docs/api/java/util/Objects.html#equals-java.lang.Object-java.lang.Object-

jkbelcher: Ha, excellent.

mcslee:
Okay some significant surgery done on this one here:
https://github.com/heronarts/LX/pull/111

I nuked the allowed channels thing, I feel quite strongly that's not the right way to go about it, and having that pattern available will lay groundwork for a lot of confusion. I think there are two ways to go about this scenario:
- UX configurability with visual feedback of what channels are controlled
- Or the user should do custom MIDI mapping if they want gaps and things

It's on my roadmap to allow collapsing group channels, like Ableton does, and which control surfaces respect. This would help a lot. I also want to add visual feedback in the UI of what's being addressed, again like Ableton.

e.g. this yellow outline shows you where your control surface is
Screen Shot 2024-05-14 at 9 55 25 AM

Separately, made `LXMidiParameterControl` instead of `LXPickupParameter` - that's really a one-way control adapter and I don't think it belongs in the core parameter package, also doesn't need to conform to the `LXParameter` interface itself. It's for writing values, not for reading/listening. I added value scaling as an optional control mode, like Ableton has. This could be generally useful elsewhere, I'll probably add it to the APC40mk2 now.

I didn't like the Bank A/B/C/D thing... feels an odd hard-coding, so I made a DiscreteParameter on `MixerSurface` which tracks the number of available mixer channels. Any surface could just use this. There are helper methods to increment/decrement by 1 or by the bank size.

Think that's the main stuff changed!

jkbelcher:
> I nuked the allowed channels thing,

NP, that was a band aid to allow visible channel customizing through overrides.  Agree wouldn't want to open it up and take it away later.

> LXMidiParameterControl instead of LXPickupParameter 

This looks good, thanks.  It's cleaner to follow and `pickupDirection` is a good add.

> I didn't like the Bank A/B/C/D thing... feels an odd hard-coding,

I went off trail here.  Was somehow thinking that when using two surfaces for 16 channels, that the 2nd controller could stay locked to Bank2 even if a new project was opened with fewer than 8 channels.  Forgot that the surface settings would be reset by the new project file regardless.  I guess if you're using one of these dedicated to CH9-16 that it will need to be set for every project file after at least 9 channels have been created.

mcslee:
> Forgot that the surface settings would be reset by the new project file regardless. I guess if you're using one of these dedicated to CH9-16 that it will need to be set for every project file after at least 9 channels have been created.

Yeah open question here whether some surface settings should be global/workspace rather than project-scoped. Ableton seems to do it globally. There are reasonable arguments either way.

> I guess if you're using one of these dedicated to CH9-16 that it will need to be set for every project file after at least 9 channels have been created.

I think once I have a box drawing in the UI to indicate where the `MixerSurface` is focused, we should enable moving it with buttons on the surface and I think that'll make performing this edit very natural, e.g. clear visual feedback where things are loaded rather than having to look at a setting and push/pull a random fader to confirm that it's lined up with what you think it's lined up with.

PR #112, opened 2024-08-14
Add retrieval of LXModel children by metadata key

jkbelcher:
I'd like to search for descendants of an LXModel that contain a metadata key.  Added methods here that mirror the methods `children(tag)` and `sub(tag)`.

### DIFF
--- a/src/main/java/heronarts/lx/model/LXModel.java
+++ b/src/main/java/heronarts/lx/model/LXModel.java
@@ -176,6 +176,12 @@ public class LXModel extends LXNormalizationBounds implements LXSerializable {
   private final Map<String, List<LXModel>> subDict =
     new HashMap<String, List<LXModel>>();
 
+  private final Map<String, List<LXModel>> metaChildDict =
+    new HashMap<String, List<LXModel>>();
+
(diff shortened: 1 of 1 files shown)


mcslee:
Gonna veto this one unfortunately. I think this is just overloading/confusing the functionality of tags. Tags are already non-unique string-key identifiers that create model-based indexing. Metadata is for retrieval of key-pair values.

Indexing every metadata key into the model structure is a potentially nontrivial increase in memory overhead for the LXModel tree, especially when you consider that parameters of LXF files are automatically included in the metadata fields, as well as things like geometry properties of grids. Definitely no need to be indexing all these model structures for having fields like "width" and "height" in metadata.

Is there some reason why you can't just apply an extra tag for what you are looking to do here? Seems like if you're already putting some key into "meta" you can also put it in tags?

jkbelcher:
Oh, definitely didn't realize metadata fields contained anything other than user-specified metadata in LXFs.

No problem on this.  I can iterate over the model tree and look for my target metadata.  Yes, adding a paired tag to find it is a good idea.  Likely will go with that option for speed now knowing the volume of metadata and tags are already indexed.

PR #113, opened 2024-09-30
MultiMonitor updates last value for all parameters

jkbelcher:
It looks like `MultiMonitor` should call `monitor.changed()` for all parameters even after a change is found, in case multiple parameters were modified in one frame.

Otherwise for example if two parameters were changed in one frame, I think `MultiMonitor.changed()` was returning `true` for two frames in a row.


mcslee:
Durf, yup of course. Nice catch.

PR #114, opened 2024-10-02
Prevent washout accumulation in BlurEffect

jkbelcher:
Have felt in the past like BlurEffect too easily turned into white.  I think this was the culprit.

To reproduce:
- Set a `SolidPattern` on red with Brightness = 5%.  It is almost not visible.
- Add a `BlurEffect`: Level=100, Decay=1s, Factor=50%.
- Enable the `BlurEffect`.  The blurBuffer rapidly accumulates to full red, 100% brightness.

I'm thinking the intention is for new colors to have an "overwrite" effect e.g. max(blurBuffer, new) but not an additive effect.  If the pattern jumped around between colors this was easily creating white.  Seems better now in local testing.

Added a `return` when disabled to save the math overhead, since the blurBuffer is reset when re-enabled.

### DIFF
--- a/src/main/java/heronarts/lx/effect/BlurEffect.java
+++ b/src/main/java/heronarts/lx/effect/BlurEffect.java
@@ -89,6 +89,10 @@ public class BlurEffect extends LXEffect {
 
   @Override
   public void run(double deltaMs, double amount) {
+    if (!isEnabled()) {
+      return;
+    }
+
     final int blurAlpha = (int) (LXColor.BLEND_ALPHA_FULL * amount * this.level.getValue());
     final int[] blurColors = this.blurBuffer.getArray();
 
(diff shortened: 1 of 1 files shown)


src/main/java/heronarts/lx/effect/BlurEffect.java
```diff
@@ -89,6 +89,10 @@ protected void onEnable() {
 
   @Override
   public void run(double deltaMs, double amount) {
+    if (!isEnabled()) {
+      return;
+    }
+
```
mcslee: This shouldn't be there. Effects automatically damp themselves when off and `run()` is not called when that's finished. But it's possible for `isEnabled()` to be false while `amount` is still positive as the effect fades out.

src/main/java/heronarts/lx/effect/BlurEffect.java
```diff
@@ -100,7 +104,7 @@ public void run(double deltaMs, double amount) {
       // Apply exponential decay to the blur
       blurColors[i] = LXColor.multiply(blurColors[i], decayColor, LXColor.BLEND_ALPHA_FULL);
       // Add the new blur buffer frame
-      blurColors[i] = LXColor.add(blurColors[i], this.colors[i], LXColor.BLEND_ALPHA_FULL);
+      blurColors[i] = LXColor.lightest(blurColors[i], this.colors[i], LXColor.BLEND_ALPHA_FULL);
```
mcslee, line 103: Yup, annoyed I got this wrong initially and that it's now a "breaking" change (in the sense that behavior will look different for same settings in a saved project).

But I agree this is definitely how it *should* behave. Def shouldn't accumulate more brightness than is actually present in the source material.

Just nuke the `isEnabled()` check and we're good.

mcslee (changes requested):

src/main/java/heronarts/lx/effect/BlurEffect.java
jkbelcher: Makes sense, thanks.  Changed back.

PR #115, opened 2024-10-05
Increment BoundedParameter relative to base

jkbelcher:
Calls to increment a BoundedParameter should modify the value relative to base value, not modulated value!

This was apparent if you called `parameter.incrementValue(0)` on a modulated parameter every frame... the modulated value became the new base value.


mcslee:
Oof, thx for catching!

PR #120, opened 2024-11-27
Dispose mixer surface field from APC surfaces

jkbelcher:
Showed up as a stranded listener on numscenes.


mcslee:
Nice catch, thx for that

PR #125, opened 2025-02-17
Remove javadoc invalid characters

jkbelcher:


mcslee:
FYI if you want those in there I think you're meant to do the html-entity thing, `&gt;` and `&lt;`

PR #130, opened 2025-03-15
Wrap hue returned by helper method when color cycling with an offset

jkbelcher:


mcslee:
Yeah better safe than sorry. Think I'd "micro-optimized" this out since some callers were doing additional math on it and then modding again themselves. But that's dumb. Any caller using this value repeatedly should call it once outside of a loop, not on a per-pixel basis.

PR #137, opened 2025-07-04
reproduce LXF parsing bug: "true&&true"

andrewlook:
Reproducing a small evaluator bug: `true&&true` (bug is also present for `true && true`). It gives this stack trace:

```
java.lang.IllegalArgumentException: Cannot evaluate empty expression:
	at heronarts.lx.structure.JsonFixture._evaluateExpression(JsonFixture.java:1285)
	at heronarts.lx.structure.JsonFixture._evaluateBooleanExpression(JsonFixture.java:1069)
	at heronarts.lx.structure.JsonFixture._evaluateExpression(JsonFixture.java:1199)
```

I'm not sure how you feel about JUnit, but this was the easiest way for me to share a directly reproducible version of the bug without loading up the full Chromatik app and sharing a fixture file.

I'm also not sure how you feel about structuring code for testability - I noticed that none of the evaluate methods used any object state, so I made them `static` (and package-protected rather than `private`) just to make it easier to isolate and test that one piece of functionality.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,6 +69,12 @@
 		    <artifactId>slf4j-simple</artifactId>
 		    <version>2.0.16</version>
 		</dependency>
+		<dependency>
+		    <groupId>org.junit.jupiter</groupId>
+		    <artifactId>junit-jupiter</artifactId>
+		    <version>5.10.1</version>
+		    <scope>test</scope>
+		</dependency>
 	</dependencies>
 
 	<build>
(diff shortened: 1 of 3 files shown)


mcslee:
Great - thanks for that!

Merged this and then went ahead and cleaved out the stateless expression stuff into its own class, doesn't need to be cluttering up `LXFixture`: https://github.com/heronarts/LX/commit/3e75f50efe709874004d678167e2238b9ebbe4d0

PR #140, opened 2025-07-16
Loop aux cue channel even if auto-muted

jkbelcher:
When a channel is aux-cued it should get looped for the preview.


mcslee:
Nice catch, thanks.

PR #141, opened 2025-07-28
LXOscConnection.filter: allow comma-delimited list

andrewlook:
Adding support for multiple filter prefixes. For example, `/lx/tempo,/lx/palette/swatch/color/`.

Only thing I was curious to get your take on: I'm not sure how hot the inner loops of `LXOscEngine` are, and whether looping an array comparison is something that we'd want to avoid for performance reasons. It seems like this would only affect people who are using filters at all, but wouldn't impact OSC outputs that aren't using filters.

### DIFF
--- a/src/main/java/heronarts/lx/osc/LXOscConnection.java
+++ b/src/main/java/heronarts/lx/osc/LXOscConnection.java
@@ -68,6 +68,8 @@ public abstract class LXOscConnection extends LXComponent {
     new StringParameter("Filter", "/lx")
     .setDescription("Filter OSC messages on matching prefix");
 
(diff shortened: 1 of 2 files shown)


mcslee:
Yeah that's not too hot a loop, and as you say, there's a check for whether filtering is even enabled, it's usually not and will just skip entirely.

I went ahead and did the TODO:
https://github.com/heronarts/LX/commit/2427a2deda7feaa3ae79ea29560a8580b5406633

Switched to just re-using a `final ArrayList` for the parsed filtered. Definitely worth doing that trimming and skipping empty strings, an unnoticed trailing space here could cause someone great hours of headache.

PR #143, opened 2025-08-03
fix: LXDeviceComponent.getModelView NPE

andrewlook:
We [noticed](https://github.com/titanicsend/LXStudio-TE/pull/688#issuecomment-3146950180) that `getParent()` is sometimes null when this method is called, and the switch statement throws an exception if called with a null value. Adding a check first and returning the same value as the default case.

### DIFF
--- a/src/main/java/heronarts/lx/LXDeviceComponent.java
+++ b/src/main/java/heronarts/lx/LXDeviceComponent.java
@@ -18,9 +18,22 @@
 
 package heronarts.lx;
 
+import java.io.IOException;
+import java.io.PrintWriter;
+import java.io.StringWriter;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.concurrent.CopyOnWriteArraySet;
+
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
+
 import heronarts.lx.midi.LXMidiListener;
 import heronarts.lx.midi.LXShortMessage;
 import heronarts.lx.midi.MidiFilterParameter;
@@ -41,17 +54,6 @@ import heronarts.lx.parameter.MutableParameter;
 import heronarts.lx.pattern.LXPattern;
 import heronarts.lx.structure.view.LXViewDefinition;
 import heronarts.lx.structure.view.LXViewEngine;
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.CopyOnWriteArraySet;
 
 /**
  * A component which may have its own scoped user-level modulators. The concrete subclasses
--- a/src/main/java/heronarts/lx/LXDeviceComponent.java
+++ b/src/main/java/heronarts/lx/LXDeviceComponent.java
@@ -18,22 +18,9 @@
 
 package heronarts.lx;
 
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.io.StringWriter;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.concurrent.CopyOnWriteArraySet;
-
 import com.google.gson.JsonArray;
 import com.google.gson.JsonElement;
 import com.google.gson.JsonObject;
-
 import heronarts.lx.midi.LXMidiListener;
 import heronarts.lx.midi.LXShortMessage;
 import heronarts.lx.midi.MidiFilterParameter;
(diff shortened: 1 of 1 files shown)


mcslee:
This feels like it’s masking another error. Don’t think getParent()
*should* ever be null when this is called.

Do you have a stack trace of that happening? Would love to track down how
or why a pattern/effect that has not been added to the lx hierarchy is
trying to call this method.

Le dim. 3 août 2025 à 21:02, Andrew Look ***@***.***> a
écrit :

andrewlook:
Sure - heres the full stack trace:

```
heronarts.lx.command.LXCommand$InvalidCommandException: SimplifyEffect could not be loaded. Check that all required content files are present and constructor is public.
	at heronarts.lx.command.LXCommand$Channel$AddEffect.perform(LXCommand.java:1391)
	at heronarts.lx.command.LXCommandEngine.perform(LXCommandEngine.java:62)
	at heronarts.lx.studio.ui.UIContentPicker$UIContentList$DeviceItem.onActivate(UIContentPicker.java:138)
	at heronarts.glx.ui.component.UIItemList$Impl.activate(UIItemList.java:681)
	at heronarts.glx.ui.component.UIItemList$Impl.onMousePressed(UIItemList.java:1029)
	at heronarts.glx.ui.component.UIItemList$ScrollList.onMousePressed(UIItemList.java:1528)
	at heronarts.glx.ui.UIObject.mousePressed(UIObject.java:697)
	at heronarts.glx.ui.UIObject.mousePressed(UIObject.java:671)
	at heronarts.glx.ui.UIObject.mousePressed(UIObject.java:671)
	at heronarts.glx.ui.UI2dScrollContext.mousePressed(UI2dScrollContext.java:286)
	at heronarts.glx.ui.UI$UIContextOverlay.mousePressed(UI.java:547)
	at heronarts.glx.ui.UIObject.mousePressed(UIObject.java:671)
	at heronarts.glx.ui.UI$UIRoot.mousePressed(UI.java:155)
	at heronarts.glx.ui.UI.mouseEvent(UI.java:1081)
	at heronarts.glx.InputDispatch.dispatch(InputDispatch.java:285)
	at heronarts.lx.LXEngine._processInputEvents(LXEngine.java:1051)
	at heronarts.lx.LXEngine.processInputEvents(LXEngine.java:1027)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)
	at java.base/java.util.concurrent.FutureTask.runAndReset(FutureTask.java:358)
	at java.base/java.util.concurrent.ScheduledThreadPoolExecutor$ScheduledFutureTask.run(ScheduledThreadPoolExecutor.java:305)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)
	at java.base/java.lang.Thread.run(Thread.java:1583)
Caused by: heronarts.lx.LX$InstantiationException: SimplifyEffect could not be loaded. Check that all required content files are present and constructor is public.
	at heronarts.lx.LX.instantiateComponent(LX.java:1462)
	at heronarts.lx.LX.instantiateEffect(LX.java:1495)
	at heronarts.lx.command.LXCommand$Channel$AddEffect.perform(LXCommand.java:1383)
	... 22 more
Caused by: java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:72)
	at java.base/java.lang.reflect.Constructor.newInstanceWithCaller(Constructor.java:502)
	at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:486)
	at heronarts.lx.LX.instantiateComponent(LX.java:1456)
	... 24 more
Caused by: java.lang.NullPointerException
	at java.base/java.util.Objects.requireNonNull(Objects.java:233)
	at heronarts.lx.LXDeviceComponent.getModelView(LXDeviceComponent.java:208)
	at titanicsend.effect.SimplifyEffect.refreshModels(SimplifyEffect.java:180)
	at titanicsend.effect.SimplifyEffect.<init>(SimplifyEffect.java:161)
	at java.base/jdk.internal.reflect.DirectConstructorHandleAccessor.newInstance(DirectConstructorHandleAccessor.java:62)
	... 27 more

```


It happens when I'm loading @jkbelcher's [SimplifyEffect](https://github.com/titanicsend/LXStudio-TE/blob/main/te-app/src/main/java/titanicsend/effect/SimplifyEffect.java), which finds the current view and enumerates the underlying fixtures (so it can average/blend the colors at the fixture level).

I think the tricky part comes when this wants to load up the model metadata in the Effect's constructor, and we also hit this when one of the relevant parameters is modified during restore from file. More discussion in the TE PR here: https://github.com/titanicsend/LXStudio-TE/pull/688#issuecomment-3146950180

mcslee:
Got it, yeah doing that at constructor time is a no go, and should also be
careful from the parameter listener. This change could introduce other
problems there because your null check may return a view that is *not* the
view that’ll be rendered by the next animation loop.

Better to set a dirty flag and update stuff in the next run() pass when you
know you’re in the proper animation context. Would that be a manageable
change to the effect implementation?

Le dim. 3 août 2025 à 21:39, Andrew Look ***@***.***> a
écrit :

mcslee:
Actually apologies, realizing I broke this recently when adding pattern
racks and it would have done what you’re now doing before. So we should
just preserve that backwards compatibility.

My change:
https://github.com/heronarts/LX/commit/82a406249e707ce000d588007d7dd29c13ffabf4

I think you can actually toss a “case null” in there but the explicit check
is also fine. Will merge when back online.

jkbelcher:
And yes this should have used a flag to refresh on the next run.  Thanks for pointing that out.  Fixed here:
[d5284db](https://github.com/titanicsend/LXStudio-TE/pull/695/commits/d5284db7664f550eb4065984a7dc9801f50be42f)

PR #144, opened 2025-08-05
Always save TriggerParameter as false

jkbelcher:
I think there's a code path where a `TriggerParameter` could fire off a serialization of its parent, and the `TriggerParameter` value would get stored as true.  This would cause it to fire whenever the object was loaded which is undesired.


mcslee:
Yeah I think that's right, I've thought about this before. I can't think of an obvious reason this would cause bad downstream effects.

PR #145, opened 2025-08-10
Fix null checks on OscEngine and Tempo

jkbelcher:
Caught a few null checks that were checking the string and not the listener object.


mcslee:
Durf, full on amateur hour material here. Thanks for catching.

PR #147, opened 2025-10-08
Refinements to registry tags

jkbelcher:


mcslee:
Made a few changes on top of that:
https://github.com/heronarts/LX/commit/029d4c8f25ad960aad67eac236028a0a555f869a

Don't love maintaining separate `mutableTags` and `tags` - just gonna wrap with `Collections.unmodifiableList` when `getTags` is called (not often), very cheap/lightweight wrapper to avoid risk of data inconsistency.

Also changed init to use the proper `addPattern`/`addEffect` methods, so the default tags are added that way. Avoid the risk of maintaining logic for component registration in two places. Got rid of the fugly `static` block for List initialization in the process.

